Issue #5, opened 2017-06-15
Create 'include' tag


sebastienros:
Define a local and a global FileProvider.

hishamco:
Is the intend of this to bring something similar to `include` in Classic ASP?

PR #7, opened 2017-06-21
Add include tag

hishamco:


hishamco:
Address issue #5

sebastienros:
Thanks Hisham, almost perfect, if only you could move this tag to the main grammar though? I created the issue before working on the view engine, this should also work in the template engine, not just the views. But all you did would just be moved in the Fluid project.

The only difference is that we need to introduce an IFileProvider in the template context instead. And maybe another one which would be `Global`, it would check if the local one is not null, otherwise fallback to the global one. It can be null too by default to prevent includes by default, for security.

Ping me if you need any more info. Thanks again.

After that you still haven't created any PR on one of my other projects ;)

sebastienros:
Also, look at the documentation for the features it supports: https://help.shopify.com/themes/liquid/tags/theme-tags#include

You can pass variables values that will be available for the sub template.
And the template name doesn't include the extension.
Also if you use `for` the value next to it will be a variable of the same name inside the template.

sebastienros:
Wait, this might just be in the shopify theme engine actually, which would mean it's fine to have it in the views only. Give me some more time to be sure.

sebastienros:
Another requirement, based on this documentation: http://www.rubydoc.info/github/Shopify/liquid/Liquid/Include

The `for` keyword will enumerate the expression and call the template for each value of the enumeration like a `with` would do.

sebastienros:
Final word, move 'include' to Fluid, this is a standard tag. Create a default FileProvider instance that will be a "BlankFileProvider" that just returns nothing.

hishamco:
Just I have two points about this:

- It would be nice if we use the built-in `NullFileProvider` instead
- While we no longer depend on `Fluid.MvcViewEngine` I used the value of the constant `ViewPath` directly

Fluid/Ast/IncludeStatement.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+            Path = path;
+        }
+
+        public Expression Path { get; }
+
+        public override async Task<Completion> WriteToAsync(TextWriter writer, TextEncoder encoder, TemplateContext context)
+        {
+            var relativeLayoutPath = (await Path.EvaluateAsync(context)).ToStringValue();
+            var currentViewPath = context.AmbientValues["ViewPath"] as string;
```
sebastienros: I think it's not necessary in this case, and actually wrong. Because for an include the file will be specified from the root of the file provider. It is different than in the view engine where we have more conventions, like requiring to have the views provided by the LocationExpander.

For instance, if the dev wants the partials in a `/partials` folder then he will provide a file provider with a root there.

Fluid/Ast/IncludeStatement.cs
```diff
@@ -0,0 +1,40 @@
...
+        public IncludeStatement(Expression path)
+        {
+            Path = path;
+        }
+
+        public Expression Path { get; }
+
+        public override async Task<Completion> WriteToAsync(TextWriter writer, TextEncoder encoder, TemplateContext context)
+        {
+            var relativeLayoutPath = (await Path.EvaluateAsync(context)).ToStringValue();
```
sebastienros: rename to `relativePath` only

Fluid/Ast/IncludeStatement.cs
```diff
@@ -0,0 +1,40 @@
...
+            var currentViewPath = context.AmbientValues["ViewPath"] as string;
+            var currentDirectory = System.IO.Path.GetDirectoryName(currentViewPath);
+            var partialPath = System.IO.Path.Combine(currentDirectory, relativeLayoutPath);
+            var fileProvider = context.AmbientValues["FileProvider"] as IFileProvider ?? _blankFileProvider;
```
sebastienros: Not from the ambient values, but from a property on the TemplateContext.

hishamco: Is that mean I should add `FileProvider` on the `TemplateContext`?

Fluid/Ast/IncludeStatement.cs
hishamco: So the `IFileProvider` will get the absolute path from `relativePath`. Correct me if I'm wrong

Fluid/Ast/IncludeStatement.cs
sebastienros: Yes, and also a GlobalFileProvider which is never null. The one in TemplateContext can be null.

sebastienros: Later I will refactor that thing where the same property can be in TC or in a global member. It will be simpler.

Fluid/Ast/IncludeStatement.cs
sebastienros: Good point, call it `path` ;) or `absolutePath`

Fluid/Ast/IncludeStatement.cs
hishamco: Why we need the `GlobalFileProvider`?

sebastienros: to provide a default application wide, and one in templatecontext if each execution needs a different value.

sebastienros: Think multi-tenancy, where themes will be looking for partials in different folders

Fluid.MvcViewEngine/FluidRendering.cs
```diff
@@ -62,6 +62,9 @@ static FluidRendering()
             context.AmbientValues[ViewPath] = path;
             context.AmbientValues.Add("Sections", new Dictionary<string, IList<Statement>>());
 
+            var viewsPath = Path.Combine(_hostingEnvironment.ContentRootPath, "Views");
+            context.FileProvider = new PhysicalFileProvider(viewsPath);
```
hishamco: I used `PhysicalFileProvider` which I dislike because the `fileProvider` will point to the root and I'm not able to get the views path from the `IncludeStatment`

IMHO we need to add a property in the `Options` to let the devs specify the partials folder, it's similar to what you said before

Fluid.MvcViewEngine/FluidRendering.cs
```diff
@@ -62,6 +62,9 @@ static FluidRendering()
             context.AmbientValues[ViewPath] = path;
             context.AmbientValues.Add("Sections", new Dictionary<string, IList<Statement>>());
 
+            var viewsPath = Path.Combine(_hostingEnvironment.ContentRootPath, "Views");
```
sebastienros: Here just reuse the existing `fileProvider` variable. You can then optionally wrap it in another one that will convert the calls to something like `/Shared`. Maybe there is already a FileProvider implementation that does it.

sebastienros: And the `Shared` folder could be an option too with it by default, that would totally make sense.

Fluid.MvcViewEngine/FluidRendering.cs
sebastienros: See previous comment about it

Fluid.MvcViewEngine/FluidRendering.cs
hishamco: I already did something similar to the above line, but I see myself make the partials folder as magic string which doesn't make sense to me

Let me show you my last attempt, after that you can review it if you don't mind

hishamco: After deep thinking the mock `FileProvider` that I made in the test project is fine for the test, so I introduce a new `FileProviderWrapper` in the view engine only

sebastienros: Sounds good, could have been named "Mapper" like the middleware map function.

hishamco: `FileProviderMapper`?!!

sebastienros: Good

sebastienros: Do you sleep sometimes?

hishamco: At first glance the name is confusing little bit, but it make sense because it map from filepath to `{Partial Folder}/filepath` 😄

Fluid/Ast/IncludeStatement.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System.IO;
+using System.Text.Encodings.Web;
+using System.Threading.Tasks;
+using Microsoft.Extensions.FileProviders;
+
+namespace Fluid.Ast
+{
+    public class IncludeStatement : Statement
+    {
+        private static readonly IFileProvider _blankFileProvider = new NullFileProvider();
```
sebastienros: Remove this line, use `TemplateContext.GlobalFileProvider` that you have created.

Fluid/Ast/IncludeStatement.cs
```diff
@@ -0,0 +1,37 @@
...
+
+            using (var stream = fileInfo.CreateReadStream())
+            using (var streamReader = new StreamReader(stream))
+            {
+                partialContent = await streamReader.ReadToEndAsync();
+            }
+
+            await writer.WriteAsync(partialContent);
+
+            return Completion.Continue;
```
sebastienros: `Completion.Normal`, Continue is for when a statement informs the container loop to step to the next element.

Fluid/Ast/IncludeStatement.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            Path = path;
+        }
+
+        public Expression Path { get; }
+
+        public override async Task<Completion> WriteToAsync(TextWriter writer, TextEncoder encoder, TemplateContext context)
+        {
+            var relativePath = (await Path.EvaluateAsync(context)).ToStringValue();
+            var fileProvider = context.FileProvider ?? _blankFileProvider;
```
sebastienros: `var fileProvider = context.FileProvider ?? TemplateContext.GlobalFileProvider;`

Fluid/TemplateContext.cs
```diff
@@ -31,10 +32,14 @@ public class TemplateContext
         /// <remarks>
         /// This property should only be set by static constructores to prevent concurrency issues.
         /// </remarks>
-        public static IMemberAccessStrategy GlobalMemberAccessStrategy = new MemberAccessStrategy(); 
+        public static IMemberAccessStrategy GlobalMemberAccessStrategy = new MemberAccessStrategy();
+
+        public static IFileProvider GlobalFileProvider { get; set; }
```
sebastienros: `= new NullFileProvider()`

PR #9, opened 2017-06-21
Fix Fluid.MvcSample debug issue in VS 15.3.0 Preview 2.0

hishamco:
The project sample working fine from the CLI, nothing but when I try to debug it from the VS it shows an error message

After modifying the .csproj you will be able to do both ways


sebastienros:
It works for me though.

hishamco:
I'm getting the following message **"A project with an Output Type of Class Library cannot be started"**, while the `Fluid.MvcSample` is the startup project

PR #16, opened 2017-08-04
Adding test to ensure parsed numbers are culture insensitive

sebastienros:
Changed Irony to support custom cultures when parsing. Default culture is `InvariantCulture` which is the one used in Fluid.

Here is the corresponding change in Irony: https://github.com/sebastienros/irony/commit/4388b1fa9d79bb2d037ab6b7afd6b22fe4520aca


jtkech:
Looks good to me.

As a test, in `Orchard.Queries` of the `fluidView` branch, i've referenced `"Irony.Core" Version="1.0.7"`. Then the unit tests pass without using `InvariantNumberLiteral`.

👍

PR #25, opened 2017-10-13
Updated to Netstandard 2.0 and added VB.NET example project

TheConservativeCoder:
Hi,

I've updated fluid to netstandard 2.0 and have included an example project that mirrors your example except that it uses VB.NET instead.

### DIFF
--- a/Fluid.MvcSampleVB/Program.vb
+++ b/Fluid.MvcSampleVB/Program.vb
@@ -4,13 +4,12 @@ Imports Microsoft.AspNetCore.Hosting
 
 Module Program
     Sub Main(args As String())
-        Dim host As IWebHost = New WebHostBuilder().
+        Call New WebHostBuilder().
             UseKestrel().
             UseContentRoot(Directory.GetCurrentDirectory()).
             UseIISIntegration().
             UseStartup(Of Startup)().
-            Build()
-
-        host.Run()
+            Build().
+            Run()
     End Sub
 End Module


Fluid.MvcSampleVB/Program.vb
```diff
@@ -0,0 +1,16 @@
+Imports System
+Imports System.IO
+Imports Microsoft.AspNetCore.Hosting
+
+Module Program
+    Sub Main(args As String())
+        Dim host = New WebHostBuilder() _
```
hishamco: I don't think the line continuation is needed here

Fluid.MvcSampleVB/Startup.vb
```diff
@@ -0,0 +1,27 @@
+﻿Imports Microsoft.Extensions.DependencyInjection
```
hishamco: Sort the namespaces

Fluid.MvcSampleVB/Startup.vb
```diff
@@ -0,0 +1,27 @@
...
+Imports FluidMvcViewEngine
+Imports Microsoft.AspNetCore.Builder
+Imports Microsoft.AspNetCore.Hosting
+Imports Microsoft.Extensions.Logging
+
+Public Class Startup
+    Shared Sub New()
```
hishamco: Why the constructor is `Shared`

Fluid.MvcSampleVB/Startup.vb
```diff
@@ -0,0 +1,27 @@
...
+        TemplateContext.GlobalMemberAccessStrategy.Register(Of Person)()
+    End Sub
+
+    Public Sub ConfigureServices(services As IServiceCollection)
+        services.AddMvc() _
```
hishamco: same as previous comment

Fluid.MvcSampleVB/Program.vb
TheConservativeCoder: Do you mean that the newline format should be kept except without the `_`, or should it all be done all on one line?

hishamco: Nope, VB multi line continuation, in declaration case you can try this  
```vb
Call New WebHostBuilder().
            UseKestrel().
            UseContentRoot(Directory.GetCurrentDirectory()).
            UseIISIntegration().
            UseStartup(Of Startup)().
            Build().
            Run()
```

hishamco: Or you can use `With` statement if you do so

TheConservativeCoder: Sweet, Didn't know about the `.` on the end of a line for it to continue!

Fluid.MvcSampleVB/Program.vb
```diff
@@ -4,7 +4,12 @@ Imports Microsoft.AspNetCore.Hosting
 
 Module Program
     Sub Main(args As String())
-        Dim host = New WebHostBuilder().UseKestrel().UseContentRoot(Directory.GetCurrentDirectory()).UseIISIntegration().UseStartup(Of Startup)().Build()
+        Dim host = New WebHostBuilder().
```
hishamco: Add missing `As`

hishamco:
Thanks for your contribution @TheConservativeCoder, don't forget to squash your commits
@sebastienros hope it's fine for you

Fluid.MvcSampleVB/Program.vb
```diff
@@ -4,7 +4,7 @@ Imports Microsoft.AspNetCore.Hosting
 
 Module Program
     Sub Main(args As String())
-        Dim host = New WebHostBuilder().
+        Dim host As IWebHost = New WebHostBuilder().
```
hishamco: It's enough to use
```vb
Dim host As New WebHostBuilder()
```

TheConservativeCoder: If it is done like that I would then be unable to call `host.Run()` as it belongs to `IWebHost`?

hishamco: I see, so that's why I surprised how `New WebHostBuilder().` it works with you without using `Call` 😄 

thanks

TheConservativeCoder: ahh I see you are calling it directly inline as one. I misread your example earlier 😊

hishamco: No problem even your latest suggestion was fine
thanks again

sebastienros:
What is the point of targeting netstandard 2.0 when we can just require 1.6 ?

TheConservativeCoder:
So that it is usable with VB.NET

sebastienros:
Can we multi-target then?

TheConservativeCoder:
Not sure what you mean? Are you saying VB.NET With ASP.NET Core 2 can reference Net Standard 1.6 or Net46?

sebastienros:
just don't remove the netstandard1.6 when you add netstandard2.0, and use conditional <PropertyGroup and <ItemGroup>

hishamco:
Multi targeting is a good option as @sebastienros said before

TheConservativeCoder:
@sebastienros What you are saying is that I should enable multi targetting by setting it like this?
`<TargetFrameworks>netstandard1.6;net46;netstandard2.0</TargetFrameworks>`

I'm not majorly familar with MSBuild stuff but I also changed the Common.props file as it has `<TargetFrameworks>netstandard1.6</TargetFrameworks>` should this be left also?

TheConservativeCoder:
Closing as replaced with #27.

hishamco:
I'm not sure why you create a new PR, just add `netstandard2.0` to target frameworks

TheConservativeCoder:
@hishamco I thought it would be easier as it would then be condensed down into small commits instead of lots of little ones. Appologies if this is not the right way, I'm not used to the etiquette of doing this with git and open (Used to using TFS instead).

hishamco:
No problem I already such things on my early days of ASP.NET 5 😄 , nothing but you can add a commits, after that use `Squash` to make them single one

sebastienros:
You don't even need to squash, we do it for you when we accept the PR (same in ASP.NET)

PR #27, opened 2017-10-23
Multitarget Netstandard 2.0 and added VB.NET example project

TheConservativeCoder:
Hi Sebastien,

I have set this up to multi target netstandard 2.0 and have added an example VB.NET Web App project that mirrors your original C# one.

### DIFF
--- a/Common.props
+++ b/Common.props
@@ -5,7 +5,7 @@
     <Authors>Sebastien Ros</Authors>
     <PackageProjectUrl>https://github.com/sebastienros/fluid</PackageProjectUrl>
     <PackageLicenseUrl>https://github.com/sebastienros/fluid/blob/master/LICENSE.txt</PackageLicenseUrl>
-    <TargetFrameworks>netstandard1.6</TargetFrameworks>
+    <TargetFrameworks>netstandard1.6;netstandard2.0</TargetFrameworks>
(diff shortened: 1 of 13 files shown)


sebastienros:
Looks good. I don't think the VB sample adds any value though.

hishamco:
I'm a VB fan 😄 , I think VB in good at this point, it shows that Fluid supports both VB.NET & C#, while Razor in ASP.NET Core still 😞

sebastienros:
I beleive if it support netstandard2.0 then it supports VB ? Any Vb user would know that without having to search for a specific Vb sample.

hishamco:
> I beleive if it support netstandard2.0 then it supports VB ?

But not for the Web, while the Razor using .chtml only

> Any Vb user would know that without having to search for a specific Vb sample.

Yep, but I think it's good to let the audience recognize that Fluid is not language dependent, as what we have seen in the Razor in ASP.NET Core

IMHO leaving the sample is fine, but the final word to you

hishamco:
@sebastienros can you take the right decision about this?

sebastienros:
1- I don't care about the Vb sample, I can add a mention in the readme to say it's supported
2- It already works in a Vb application, I just tried

Do we still need to target netstandard2.0 for any reason?

hishamco:
> Do we still need to target netstandard2.0 for any reason?

I don't think so

TheConservativeCoder:
Cool, I'll delete this

Issue #28, opened 2017-10-27
include tag behavior


jeremycook:
The `include` tag only writes the contents of the included file, but does not parse the file. This varies from [Shopify's include tag implementation](http://www.rubydoc.info/github/Shopify/liquid/Liquid/Include). The `with` and `for` syntax are also not supported at this time.

At a minimum should Fluid render the contents of the included file using the existing context?

sebastienros:
All should be handled "at a minimum". If you feel like you could help, then please do so.

jeremycook:
@sebastienros The variables I have access to are `Expression path` from the `public IncludeStatement(Expression path)` constructor, and `TextWriter writer`, `TextEncoder encoder`, `TemplateContext context` via `WriteToAsync` method. It seems to me that I need an `IFluidParser` and a `IFluidTemplate` when `WriteToAsync` is called so that I can render the contents of the include file at that time. Passing these through the constructor seems wrong. So the other option is to pass it as part of the call to `WriteToAsync`.

How should I go about making an `IFluidParser` and a `IFluidTemplate` available to the `WriteToAsync` method? Or would you recommend a different approach?

sebastienros:
We might just be able to provide the parser in the ambient value property, globally, so any filter can use it if necessary.

PR #29, opened 2017-10-27
Render include file contents

jeremycook:
#28 
I registered the ambient values locally. I wasn't sure how to do it globally.

### DIFF
--- a/Fluid.MvcViewEngine/FluidRendering.cs
+++ b/Fluid.MvcViewEngine/FluidRendering.cs
@@ -23,6 +23,7 @@ namespace FluidMvcViewEngine
     {
         private const string ViewStartFilename = "_ViewStart.liquid";
         public const string ViewPath = "ViewPath";
+        private static readonly Func<IFluidTemplate> FluidTemplateFactory = () => new FluidViewTemplate();
 
         static FluidRendering()
         {
@@ -61,7 +62,7 @@ namespace FluidMvcViewEngine
             context.AmbientValues[ViewPath] = path;
             context.AmbientValues.Add("Sections", new Dictionary<string, IList<Statement>>());
             context.AmbientValues[IncludeStatement.FluidParserFactoryKey] = FluidViewTemplate.Factory;
-            context.AmbientValues[IncludeStatement.FluidTemplateKey] = new FluidViewTemplate();
+            context.AmbientValues[IncludeStatement.FluidTemplateFactoryKey] = FluidTemplateFactory;
             context.FileProvider = new FileProviderMapper(fileProvider, "Views");
 
             var body = await template.RenderAsync(context);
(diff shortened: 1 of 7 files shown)


Fluid.Tests/IncludeStatementTests.cs
```diff
@@ -50,6 +50,8 @@ public class IncludeStatementTests
             {
                 FileProvider = new TestFileProvider("Partials")
             };
+            context.AmbientValues["FluidParser"] = FluidTemplate.Factory.CreateParser();
```
sebastienros: Pass the factory only, and create the parser when necessary in the filter

jeremycook: Ok, I made this change. Hopefully it is what you are expecting.

Fluid/Ast/IncludeStatement.cs
```diff
@@ -31,15 +34,37 @@ public IncludeStatement(Expression path)
...
+                throw new ArgumentException($"The '{FluidParserFactoryKey}' key was not present in the AmbientValues dictionary.", nameof(context));
+            }
+            if (!context.AmbientValues.ContainsKey(FluidTemplateKey))
+            {
+                throw new ArgumentException($"The '{FluidTemplateKey}' key was not present in the AmbientValues dictionary.", nameof(context));
+            }
+
+            FluidParserFactory factory = (FluidParserFactory)context.AmbientValues[FluidParserFactoryKey];
+            IFluidParser parser = factory.CreateParser();
+            IFluidTemplate template = (IFluidTemplate)context.AmbientValues[FluidTemplateKey];
```
sebastienros: Can't it be created from the factory? Even if we need to add something to the factory?

jeremycook: The template isn't available from the parser factory.

sebastienros: Then just put the type in the `context.AmbientValues[FluidTemplateKey]` and call Activator.CreateInstance when you need the template instance. The goal is to prevent the template allocation everytime a filter is used.

sebastienros: And also because if you call include multiple times you want a new template instance every time.

jeremycook: My latest commit uses parser and template factories from ambient values if they are present, otherwise it uses what seems like sane defaults. So nothing is "globally" configurable per-se. I set the ambient values in the MVC renderer, but do not set them or need to in the tests.

jeremycook:
@sebastienros  Are you going to squash this? I can reduce it to one commit if you like. (I don't have too much experience contributing like this on GitHub. So I don't know what's preferred.)

hishamco:
@jeremycook the commits will be one before merge automatically

Fluid.MvcViewEngine/FluidRendering.cs
```diff
@@ -71,17 +74,18 @@ static FluidRendering()
...
                 var layoutTemplate = ParseLiquidFile((string)layoutPath, fileProvider, false);
 
-                return await layoutTemplate.RenderAsync(context); 
+                return await layoutTemplate.RenderAsync(context);
             }
 
             return body;
         }
 
+
```
hishamco: Remove the extra line

hishamco:
It would be nice if you add a unit tests too

sebastienros:
Next task is to implement `for` and `with`.

@hishamco sorry, I didn't let you the time to comment

hishamco:
no problem @sebastienros

Fluid.MvcViewEngine/FluidRendering.cs
jeremycook: I did this as part of a follow up commit. Thanks for taking a look.

Issue #30, opened 2017-10-27
Implement include options


sebastienros:
https://help.shopify.com/themes/liquid/tags/theme-tags#include

hishamco:
We need to add `With` Am I right?

hishamco:
Let me check the link if I can do it ..

sebastienros:
`with`, `for` and the inline variables too

sebastienros:
It can be done in multiple steps

hishamco:
I didn't see the `for` in the link that you provide

hishamco:
May I start to support `With` for inline variable assignment

sebastienros:
the easiest should be the inline variables, but do what you prefer

sebastienros:
I say it's easier because it doesn't require to change the tag and support the `with` keyword. But you have created tags by the past, so you know how it works.

hishamco:
According what I saw the `With` basically equivalent to `assign` during including the template

hishamco:
Yep, now I see the inline variables
```
{% include 'snippet', my_variable: 'apples', my_other_variable: 'oranges' %}
```
I will start on it soon

sebastienros:
Does the current context need to be cloned and get these new variables? Or is it a brand new one?

sebastienros:
> When you include a snippet, the code inside it will have access to the variables within its parent template.

So the solution is to create a new context, and give the current one as the `parent` context. I think I implemented it this way when I created the contexts. If I remember correctly.

hishamco:
Let me check in my way ..

sebastienros:
Ah, `EnterScope` and `LeaveScope`.

hishamco:
@sebastienros could you please review the PRs related to this

hishamco:
@sebastienros this need to be closed

PR #32, opened 2017-10-28
Support inline variable assignment in Include tag

hishamco:
This is a part of issue #30

### DIFF
--- a/Fluid/Ast/IncludeStatement.cs
+++ b/Fluid/Ast/IncludeStatement.cs
@@ -22,12 +22,15 @@ namespace Fluid.Ast
         public const string FluidTemplateFactoryKey = "FluidTemplateFactory";
         public const string ViewExtension = ".liquid";
 
-        public IncludeStatement(Expression path)
+        public IncludeStatement(Expression path, IList<AssignStatement> assignStatements)
         {
             Path = path;
+            AssignStatements = assignStatements;
         }
 
         public Expression Path { get; }
+
+        public IList<AssignStatement> AssignStatements { get; }
 
         public override async Task<Completion> WriteToAsync(TextWriter writer, TextEncoder encoder, TemplateContext context)
         {
@@ -43,6 +46,14 @@ namespace Fluid.Ast
             if (!fileInfo.Exists)
             {
                 throw new FileNotFoundException(relativePath);
+            }
+
+            if (AssignStatements != null)
+            {
+                foreach (var assignStatement in AssignStatements)
+                {
+                    await assignStatement.WriteToAsync(writer, encoder, context);
+                }
             }
 
             using (var stream = fileInfo.CreateReadStream())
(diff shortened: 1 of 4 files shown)


hishamco:
@sebastienros  is passing the assignment statements into the include is a good idea, or there is way to add them to the context, something similar to `TemplateContext.GlobalScope`

Fluid/DefaultFluidParser.cs
```diff
@@ -893,7 +901,28 @@ public static FilterArgument BuildFilterArgument(ParseTreeNode node)
...
+        {
+            var identifier = tag.ChildNodes[0].Token.ValueString;
+            var value = BuildTermExpression(tag.ChildNodes[1]);
+
+            return new AssignStatement(identifier, value);
+        }
+
         #endregion
+
+        private static void Traverse(ParseTreeNode tag, ref IList<AssignStatement> statements)
```
sebastienros: Why `ref`?

sebastienros: What is recursive in include assignments?

hishamco: > Why ref?

The `ref` needed to keep track the `AssignStatement`s during the traversing, I could make as global field instead and remove from here .. I think this a better choice
> What is recursive in include assignments?

The recursive needed to visit all the node in the assignment tree

Fluid/Ast/IncludeStatement.cs
```diff
@@ -45,6 +48,14 @@ public IncludeStatement(Expression path)
                 throw new FileNotFoundException(relativePath);
             }
 
+            if (AssignStatements != null)
+            {
+                foreach (var assignStatement in AssignStatements)
```
sebastienros: I don't think these values should be visible after the include statement is executed. Probably necessary to call `context.EnterScope` and `content.LeaveScope` after.

Fluid/Ast/IncludeStatement.cs
```diff
@@ -50,10 +50,14 @@ public IncludeStatement(Expression path, IList<AssignStatement> assignStatements
 
             if (AssignStatements != null)
             {
+                context.EnterChildScope();
```
sebastienros: You need to grab the result and pass it to `assignStatement.WriteToAsync`

hishamco: I didn't get your point, what I understand is to make the assignments executed within the scope, similar to https://github.com/sebastienros/fluid/blob/dev/Fluid/Ast/IncludeStatement.cs#L51-L65

Correct me if I'm wrong

sebastienros: I was wrong ;)

sebastienros: I just didn't remember what I had done :blush:

hishamco: 😄

hishamco:
LGTM, when need to rebase to merge both of the PRs

thanks @sebastienros

PR #33, opened 2017-10-28
Add 'with' option to Include tag

hishamco:
This a second part of the issue #30

### DIFF
--- a/Fluid/Ast/IncludeStatement.cs
+++ b/Fluid/Ast/IncludeStatement.cs
@@ -53,7 +53,7 @@ namespace Fluid.Ast
 
             if (With != null)
             {
-                var identifier = (await Path.EvaluateAsync(context)).ToStringValue();
+                var identifier = System.IO.Path.GetFileNameWithoutExtension(relativePath);
 
                 await new AssignStatement(identifier, With).WriteToAsync(writer, encoder, context);
             }
--- a/Fluid/Ast/IncludeStatement.cs
+++ b/Fluid/Ast/IncludeStatement.cs
@@ -22,15 +22,18 @@ namespace Fluid.Ast
         public const string FluidTemplateFactoryKey = "FluidTemplateFactory";
         public const string ViewExtension = ".liquid";
 
-        public IncludeStatement(Expression path, IList<AssignStatement> assignStatements = null)
+        public IncludeStatement(Expression path, Expression with = null, IList<AssignStatement> assignStatements = null)
         {
             Path = path;
+            With = with;
             AssignStatements = assignStatements;
         }
 
         public Expression Path { get; }
 
         public IList<AssignStatement> AssignStatements { get; }
+
+        public Expression With { get; }
 
         public override async Task<Completion> WriteToAsync(TextWriter writer, TextEncoder encoder, TemplateContext context)
         {
@@ -51,7 +54,7 @@ namespace Fluid.Ast
             if (AssignStatements != null)
             {
                 context.EnterChildScope();
-                
+
                 try
                 {
                     foreach (var assignStatement in AssignStatements)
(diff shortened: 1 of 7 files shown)


hishamco:
@sebastienros this is straightforward, so after I got your comments about the PR #32, I can rebase to include this change

Fluid/Ast/IncludeStatement.cs
```diff
@@ -22,13 +22,16 @@ public class IncludeStatement : Statement
...
+            Value = value;
         }
 
         public Expression Path { get; }
 
+        public Expression Value { get; }
+
         public override async Task<Completion> WriteToAsync(TextWriter writer, TextEncoder encoder, TemplateContext context)
         {
             var relativePath = (await Path.EvaluateAsync(context)).ToStringValue();
```
sebastienros: Maybe save the value at this point, and not call `Path.EvaluateAsync` again ?

Fluid/Ast/IncludeStatement.cs
```diff
@@ -22,13 +22,16 @@ public class IncludeStatement : Statement
...
-        public IncludeStatement(Expression path)
+        public IncludeStatement(Expression path, Expression value = null)
         {
             Path = path;
+            Value = value;
         }
 
         public Expression Path { get; }
 
+        public Expression Value { get; }
```
sebastienros: Should it be named `With` to make it obvious what kind of value it is?

Fluid/DefaultFluidParser.cs
```diff
@@ -527,7 +527,13 @@ public static DecrementStatement BuildDecrementStatement(ParseTreeNode tag)
         public static IncludeStatement BuildIncludeStatement(ParseTreeNode tag)
         {
             var pathExpression = BuildTermExpression(tag.ChildNodes[0]);
-            return new IncludeStatement(pathExpression);
+            Expression valueExpression = null;
```
sebastienros: `withExpression` ?

Fluid/Ast/IncludeStatement.cs
hishamco: It's similar to `Path`?!!

Fluid/Ast/IncludeStatement.cs
```diff
@@ -45,6 +48,14 @@ public IncludeStatement(Expression path)
                 throw new FileNotFoundException(relativePath);
             }
 
+            if (Value != null)
+            {
+                var identifier = (await Path.EvaluateAsync(context)).ToStringValue();
+
+                await new AssignStatement(identifier, Value).
+                    WriteToAsync(writer, encoder, context);
```
sebastienros: This should be evaluated on `childScope` instead of `context`

Fluid/Ast/IncludeStatement.cs
hishamco: But we need the value of the context!!

sebastienros: If I change the code and the tests pass, will you trust me ?

hishamco: Be free to change the code if you have a better solution

> will you trust me ?

I'm always trust you 😄

hishamco:
I see, because I surprised when you told me grap the result of `context.EnterChildScope()` and pass it to the `WriteToAsync` , coz the scope can't be converted to context 😄

sebastienros:
Do you want me to merge one PR and you'll update the other one ?

hishamco:
Yep, that's what I wrote in the other PR, do it please ..

Fluid/Ast/IncludeStatement.cs
```diff
@@ -45,6 +48,21 @@ public IncludeStatement(Expression path)
...
+                var identifier = System.IO.Path.GetFileNameWithoutExtension(relativePath);
+                context.EnterChildScope();
+
+                try
+                {
+                    await new AssignStatement(identifier, With).WriteToAsync(writer, encoder, context);
+                }
+                catch
+                {
+                    context.ReleaseScope();
```
sebastienros: I think that's wrong, and the tests should have shown it.

sebastienros:
Please add a test that ensures that a variable is not set in the template that calls the include tag. For `with` and the inline parameters. This should have failed, the usage of EnterChildScope was wrong.

hishamco:
@sebastienros as you said before, there's an issue with `EnterChildScope`

hishamco:
👍

Issue #26, opened 2017-10-19
Liquid within comment executes


jeremycook:
Should Liquid within a comment still execute or should a comment prevent execution? For Shopify's implementation, code will not execute:

> comment
> Allows you to leave un-rendered code inside a Liquid template. Any text within the opening and closing comment blocks will not be output, and any Liquid code within will not be executed. https://help.shopify.com/themes/liquid/tags/theme-tags#comment

Fluid example:

    {% comment %}
    I don't get printed but this code is executed. That is, `programs` is assigned a value, and all that.
    {% assign programs = '[ {"id": 1, "iconUrl": "http://lorempixel.com/output/animals-q-c-640-480-3.jpg", "title": "One" } ]' | parse %}
    {% endcomment %}

sebastienros:
If it's executing then I assume it's a bug.

hishamco:
I can verify that `Include` tag is the only tag I ever seen that been executed inside the comment statement, I'm working on this now ..

PR #34, opened 2017-10-29
Liquid within comment should not be executed

hishamco:
Address issue #26

### DIFF
--- a/Fluid.Tests/IncludeStatementTests.cs
+++ b/Fluid.Tests/IncludeStatementTests.cs
@@ -84,7 +84,8 @@ namespace Fluid.Tests
 
             public IChangeToken Watch(string filter)
             {
-                throw new NotImplementedException();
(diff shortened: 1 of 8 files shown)


sebastienros:
Why is Include the only tag that behaves like this. Sounds like an issue with this tag.

hishamco:
This is the workaround that I come up with, that main reason behind this, because the `include` called `TryParse()` for the partial after the `{% endcomment %}` has been closed

Refer to [this](https://github.com/sebastienros/fluid/pull/29/files#diff-9de51dea2a4e6978c5be0829c3c9a06bR62)

hishamco:
IMHO workaround is acceptable, until we find a better way

sebastienros:
Can't be accepted as a workaround as it's not an issue for include tags only.

hishamco:
I saw this issue occurs only for `include` tag!!

PR #35, opened 2017-10-30
Ignore tags in raw and comment blocks

### LINKED ISSUE #26: Liquid within comment executes

sebastienros:
Fixes #26

### DIFF
--- a/Fluid.Tests/ParserTests.cs
+++ b/Fluid.Tests/ParserTests.cs
@@ -118,6 +118,16 @@ namespace Fluid.Tests
         }
 
         [Fact]
+        public void ShouldParseRawWithBlocks()
+        {
+            var statements = Parse(@"{% raw %} {%if true%} {%endif%} {% endraw %}");
+
+            Assert.Equal(1, statements.Count);
+            Assert.IsType<TextStatement>(statements.ElementAt(0));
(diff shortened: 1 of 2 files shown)


sebastienros:
Welcoming your review @hishamco

hishamco:
Can we add a test similar to [this](https://github.com/sebastienros/fluid/pull/34/files#diff-d6ec8a3529c92d09f8940f27153bfab4R450)?

hishamco:
LGTM

sebastienros:
I understand why we could add a test like you suggest, but the unit tests I wrote ensure that just the raw text is analyzed. So I think we are well covered.

Issue #36, opened 2017-10-31
Include Tag Perf


hishamco:
IMHO we need to cache parsing the included template instead of parse it each time

PR #37, opened 2017-10-31
Cache included liquid templates

hishamco:
Address issue #36

### DIFF
--- a/Fluid/Ast/IncludeStatement.cs
+++ b/Fluid/Ast/IncludeStatement.cs
@@ -3,11 +3,15 @@ using System.Collections.Generic;
 using System.IO;
 using System.Text.Encodings.Web;
 using System.Threading.Tasks;
+using Microsoft.Extensions.Caching.Memory;
+using Microsoft.Extensions.FileProviders;
 
 namespace Fluid.Ast
 {
     public class IncludeStatement : Statement
     {
+        private static readonly IMemoryCache _memoryCache = new MemoryCache(new MemoryCacheOptions());
+
         /// <summary>
         /// Override <see cref="IFluidParserFactory"/> used by a <see cref="TemplateContext"/>
         /// by setting an ambient value with a key matching the value of the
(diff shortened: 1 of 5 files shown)


hishamco:
I can ride off a lot of code by introducing:
```csharp
private class FluidViewTemplate : BaseFluidTemplate<FluidViewTemplate>
{

}
```
/cc @sebastienros

Fluid/Ast/IncludeStatement.cs
```diff
@@ -56,38 +60,56 @@ public IncludeStatement(Expression path, Expression with = null, IList<AssignSta
...
 
             return Completion.Normal;
         }
 
+        private static IFluidTemplate ParseTemplate(string path, IFileProvider fileProvider, TemplateContext context)
+        {
+            return _memoryCache.GetOrCreate(path, entry =>
+            {
+                var fileInfo = fileProvider.GetFileInfo(path);
+                entry.SlidingExpiration = TimeSpan.FromHours(1);
```
sebastienros: Static value.

hishamco: We can introduce an option to control this for both included templates and other liquid templates if it's needed

sebastienros: We could but we need to enforce there is one for security reasons. If you can make it an option too then good.

hishamco: Ok

hishamco:
The `CacheOption` for `FluidMvcViewEngine` is straightforward, I don't wanna include tag rely on the `FluidMvcViewEngine` options, coz may someone use is Fluid without `FluidMvcViewEngine`

Any idea @sebastienros ?

sebastienros:
Then don't make it an option.

hishamco:
So can I simplify the parsing using:
```csharp
private class FluidViewTemplate : BaseFluidTemplate<FluidViewTemplate>
{

}
```
or leave it as its

sebastienros:
I don't understand the relationship with this PR and FluidViewTemplate already inherits from it. Why make it private then?

hishamco:
Just an accessor, I wanna call `FluidViewTemplate.TryParse` instead of creating a new parser instance, like [this](https://github.com/sebastienros/fluid/blob/dev/Fluid/Ast/IncludeStatement.cs#L91-L115)

sebastienros:
`CreateParser` is cheap, use it, and remove the `else` cases: https://github.com/sebastienros/fluid/blob/dev/Fluid/Ast/IncludeStatement.cs#L97

hishamco:
I can't remove it because it will not compile, anyhow we can use the available `CreateParser()` without introducing something else

sebastienros:
> I can't remove it because it will not compile 

If I remove it and the tests pass will you trust me ?

hishamco:
If is that what's the need to relay on FluidParserFactoryKey?

hishamco:
Hope you will like the last change that I made

sebastienros:
If Include can be used outside of FluidMvc (which I assume is the case) then the template can be of any type (as it's customizable) hence you can use a specific type, and need to use whatever is passed in the ambient values. And not use this class anywhere in the statements.

PR #38, opened 2017-11-05
Update MiscFilters.cs

jtkech:
Maybe intended but i don't think so.


hishamco:
Nice catch

jtkech:
Sébastien makes copy paste too fast ;)

sebastienros:
Is there a bug because of that? I would have thought I'd done a test for each.

PR #43, opened 2017-11-08
Adding IndentStatement

### LINKED ISSUE #42: include tag whitespace
Hi,

When using the include tag, I would expect the contents of the included liquid file to render with the whitespace of the root file appended.
For example:
Class.liquid:

```
...
{% for col in Table.Columns %}
        {% include 'AutoProperty' %}
{% endfor %}
...
```

AutoProperty.liquid:

`public {{ col.PropertyType }} {{ col.PropertyName }} { get; set; }`

Currently outputs like this:

```
public int PrincipalId { get; set; }
public int DiagramId { get; set; }
public int Version { get; set; }
```

where I would expect this:

```
        public int PrincipalId { get; set; }
        public int DiagramId { get; set; }
        public int Version { get; set; }
```

sebastienros:
Fixes #42

### DIFF
--- a/Fluid.Tests/IncludeStatementTests.cs
+++ b/Fluid.Tests/IncludeStatementTests.cs
@@ -11,6 +11,15 @@ namespace Fluid.Tests
 {
     public class IncludeStatementTests
     {
+        MockFileProvider _fileProvider;
+
+        public IncludeStatementTests()
+        {
+            _fileProvider = new MockFileProvider()
+                .AddTextFile("_Partial.liquid", @"color: '{{ color }}' shape: '{{ shape }}'")
+                .AddTextFile("color.liquid", @"color: '{{ color }}' shape: '{{ shape }}'");
+        }
+
         [Fact]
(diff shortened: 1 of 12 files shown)


hishamco:
Can you resolve the conflict and merge? or you need to sort of revision?

hishamco:
What stop us from merging this - after resolving the conflict - ?!!

hishamco:
@sebastienros what's stop us from resolve & merge this?

sebastienros:
Probably nothing.

hishamco:
Can you please fix the conflict? we need to close all the PRs, also the issues as much as we can

hishamco:
Can we merge this?

hishamco:
Is it fixed on Parlot or closed for another reason

sebastienros:
Too out of specs, too old, won't work with new Parlot branch, easier to rethink it later

hishamco:
Ok

Issue #45, opened 2017-11-29
Include variables written to caller's scope


jeremycook:
The following is an example of this behavior. You can see that the second time I include 'template' I do not set the value of `varB` but `varB`'s value from the first include of 'template' appears. I also have a similar example of using `with`.

I'm not sure if that is how Shopify's implementation of include behaves. So I don't know that this is strictly a bug. For my purposes it causes problems, and is an unwanted side-effect.

Your thoughts on whether or not this is a bug, or how I could scope variables in a way that they don't leak from one `include` to another would be appreciated.

```c#
var folder = Path.Combine(Path.GetTempPath(), "IncludeScopeExample");
Directory.CreateDirectory(folder);
File.WriteAllText(Path.Combine(folder, "template.liquid"),
@"-----------------
varA = {{ varA }}
varB = {{ varB }}
template = {{ template }}
");

TemplateContext.GlobalFileProvider = new PhysicalFileProvider(folder);

var model = new { Firstname = "Bill", Lastname = "Gates" };
var source =
@"Hello {{ p.Firstname }} {{ p.Lastname }}

{% include 'template', varA: 'first call of varA', varB: 'first call of varB' %}
{% include 'template', varA: 'second call of varA' %}

{% include 'template' with 'third call of with' %}
{% include 'template' %}";

if (FluidTemplate.TryParse(source, out var template))
{
    var context = new TemplateContext();
    context.MemberAccessStrategy.Register(model.GetType());
    context.SetValue("p", model);

    Console.Write(template.Render(context));
}
```

Output:

```console
Hello Bill Gates

-----------------
varA = first call of varA
varB = first call of varB
template = 
-----------------
varA = second call of varA
varB = first call of varB
template = 

-----------------
varA = second call of varA
varB = first call of varB
template = third call of with
-----------------
varA = second call of varA
varB = first call of varB
template = third call of with
```

sebastienros:
I will look into some examples/tests from shopify, but it looks like a bad behavior to me. All this variables should be local to the template, and not leak on subsequent renderings. But we need proof.

sebastienros:
I can't find anything related in the tests https://github.com/Shopify/liquid/blob/9c72ccb82f4e7a1f158b59741596a473c499875e/test/integration/tags/include_tag_test.rb

It looks like the template shouldn't change the parent's variables, but that's it. From the documentation, the other values are actually defining the main context variables, so what you are seeing seems expected. We would need to check on a shopify site probably.

[...]

PR #48, opened 2017-11-30
Child and local scope handling by TemplateContext

jeremycook:
I believe this addresses the source of variables leaking into parent scope like in issue #45. This affects the `include` and `for` statements.

One thing I did was change TemplateContext.EnterChildScope from returning a `Scope` to returning `void`. This is because the TemplateContext.ReleaseScope method should be called and not the `Scope` object's `ReleaseScope`. I want to encourage callers to use TemplateContext's Release Scope method. Calling code can still technically call the Scope object's ReleaseScope method directly by accessing it via the TemplateContext.LocalScope property, but the result of doing so is undefined behavior.

This could be further refactored by introducing an interface that the Scope class would implement. The interface would not surface the `ReleaseScope` method. The TemplateContext.LocalScope property would be of this type of interface instead of Scope, and further guarding calling code calling TemplateContext.LocalScope.ReleaseScope directly.

### DIFF
--- a/Fluid.Tests/IncludeStatementTests.cs
+++ b/Fluid.Tests/IncludeStatementTests.cs
@@ -49,6 +49,7 @@ namespace Fluid.Tests
                 FileProvider = new MockFileProvider("Partials")
             };
             var expectedResult = @"Partial Content
+Partials: ''
 color: ''
 shape: ''";
 
@@ -72,6 +73,7 @@ shape: ''";
                 FileProvider = new MockFileProvider("Partials")
             };
             var expectedResult = @"Partial Content
+Partials: ''
 color: 'blue'
 shape: 'circle'";
 
@@ -91,6 +93,7 @@ shape: 'circle'";
                 FileProvider = new MockFileProvider("Partials")
             };
             var expectedResult = @"Partial Content
+Partials: ''
 color: 'blue'
 shape: ''";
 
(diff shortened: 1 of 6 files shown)


jeremycook:
I updated include tests so that they are working "as expected" now. Unfortunately there are 3 date filter tests that are failing.

sebastienros:
Would you mind implementing your suggestion by returning an `IDisposable` that would call `Release` on the `Dispose`. Then remove the try/catch statements and just do `using (context.EnterScope())`. Keep the `ReleaseScope` public, but maybe rename it to `LeaveScope` ?

jeremycook:
Here you go. The code feels quick and dirty. I use a delegate to allow the scope to reset the TemplateContext to its parent scope on dispose. It seems like a stack could be better than these recursive Scope objects. If you had two child scopes and they were disposed out of order strange things could happen. Suggestions welcome.

sebastienros:
You are right, we would need an `AsyncStatic` field in this case, just revert to the previous commit and I'll merge. Sorry to that.

jeremycook:
Should be good to go. I removed the IDisposable Scope commit.

jeremycook:
BTW what do you mean by AsyncStatic field? Is there an example of what you're thinking out in the wild?

sebastienros:
https://github.com/aspnet/Caching/blob/dev/src/Microsoft.Extensions.Caching.Memory/CacheEntryHelper.cs#L11

PR #50, opened 2017-12-01
Truncate returns the input unmodified if the string is too short or nil

jeremycook:
The truncate filter errors if the input is nil or shorter than length. This fixes that.

### DIFF
--- a/Fluid/Filters/StringFilters.cs
+++ b/Fluid/Filters/StringFilters.cs
@@ -145,14 +145,24 @@ namespace Fluid.Filters
 
         public static FluidValue Truncate(FluidValue input, FilterArguments arguments, TemplateContext context)
         {
-            var source = input.ToStringValue().Substring(0, Convert.ToInt32(arguments.At(0).ToNumberValue()));
+            string text = input.ToStringValue();
+            int length = Convert.ToInt32(arguments.At(0).ToNumberValue());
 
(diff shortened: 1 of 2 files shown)


Fluid/Filters/StringFilters.cs
```diff
@@ -145,14 +145,24 @@ public static FluidValue Strip(FluidValue input, FilterArguments arguments, Temp
 
         public static FluidValue Truncate(FluidValue input, FilterArguments arguments, TemplateContext context)
         {
-            var source = input.ToStringValue().Substring(0, Convert.ToInt32(arguments.At(0).ToNumberValue()));
+            string text = input.ToStringValue();
+            int length = Convert.ToInt32(arguments.At(0).ToNumberValue());
 
-            if (arguments.Count > 1)
+            if (text == null || text.Length <= length)
```
hishamco: Can we do it like thsi
```csharp
if (text.Length > length)
{
     text = text.Substring(0, length);
}

return text;
```

jeremycook: I think it's fine as-is.

sebastienros:
A bug fix needs a unit test that would have failed without the fix. 
In other words, thanks you for your contribution, but let me ask you for even more of your time ;)

jeremycook:
Both tests fail with the old code and pass with the new.

PR #60, opened 2018-01-28
Add ellipsis to Truncate & TruncateWords

### LINKED ISSUE #59: truncate filter doest add ellipsis by defulat
Using the truncate filter in a OC blog summary list doesn't add the ellipsis by default as specified by the liquid language reference:
https://help.shopify.com/themes/liquid/filters/string-filters#truncate

hishamco:
Fixes #59

### DIFF
--- a/Fluid.Benchmarks/Fluid.Benchmarks.csproj
+++ b/Fluid.Benchmarks/Fluid.Benchmarks.csproj
@@ -6,6 +6,12 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <Compile Remove="BenchmarkDotNet.Artifacts\**" />
+    <EmbeddedResource Remove="BenchmarkDotNet.Artifacts\**" />
+    <None Remove="BenchmarkDotNet.Artifacts\**" />
+  </ItemGroup>
+
+  <ItemGroup>
     <PackageReference Include="BenchmarkDotNet" Version="0.10.9" />
     <PackageReference Include="DotLiquid" Version="2.0.174" />
     <PackageReference Include="Liquid.NET" Version="0.9.6" />
--- a/Fluid.Tests/StringFiltersTests.cs
+++ b/Fluid.Tests/StringFiltersTests.cs
@@ -195,8 +195,10 @@ namespace Fluid.Tests
         }
 
         [Theory]
-        [InlineData("Hello World", 8, "Hello...")]
-        [InlineData("Hello", 2, "...")]
+        [InlineData("The cat came back the very next day", 13, "The cat ca...")]
+        [InlineData("Hello", 3, "...")]
+        [InlineData("Hello", 10, "Hello...")]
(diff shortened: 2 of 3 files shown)


hishamco:
After trying different scenarios with `truncate` & ellipses using http://dotliquidmarkup.org/try-online, seems our previous needs to be change, we need to compare with ellipses length instead of the input length

/cc @sebastienros

Fluid/Filters/StringFilters.cs
```diff
@@ -145,27 +145,34 @@ public static FluidValue Strip(FluidValue input, FilterArguments arguments, Temp
 
         public static FluidValue Truncate(FluidValue input, FilterArguments arguments, TemplateContext context)
         {
-            string text = input.ToStringValue();
-            int length = Convert.ToInt32(arguments.At(0).ToNumberValue());
+            var ellipsis = "...";
```
sebastienros: `const` at the class level

Fluid.Tests/StringFiltersTests.cs
```diff
@@ -199,7 +199,7 @@ public void Truncate()
         {
             var input = new StringValue("Hello World");
 
-            var arguments = new FilterArguments().Add(new NumberValue(5)).Add(new StringValue("..."));
+            var arguments = new FilterArguments().Add(new NumberValue(8));
```
sebastienros: Can you refactor to a `[Theory]` with `input`, `output` and `size` to handle the edge cases?

Fluid/Filters/StringFilters.cs
hishamco: I already did that at the beginning, but this will change at the future. We can add an ellipses as parameter

sebastienros:
@hishamco I have added more edge cases and also fixed them. GTG when I get your approval.

Fluid.Benchmarks/Fluid.Benchmarks.csproj
```diff
@@ -5,6 +5,12 @@
     <TargetFramework>netcoreapp2.0</TargetFramework>
   </PropertyGroup>
 
+  <ItemGroup>
```
hishamco: Is this needed?!!

sebastienros: Unrelated to your PR, but saw this folder shouldn't be included.

Fluid/Filters/StringFilters.cs
```diff
@@ -1,12 +1,13 @@
 ﻿using System;
-using System.Collections.Generic;
 using System.Linq;
 using Fluid.Values;
 
 namespace Fluid.Filters
 {
     public static class StringFilters
     {
+        private static StringValue Ellipsis = new StringValue("...");
```
hishamco: `static readonly` is better in this case

hishamco:
LGTM after my above comment

PR #65, opened 2018-03-01
Adds TemplateContext to property accessor delegates

paynecrl97:


paynecrl97:
Having access to the current TemplateContext allows for much more powerful member access delegates

PR #67, opened 2018-03-01
Adding async and lazy initialization support

sebastienros:
`Register` accepts `async` methods
`SetValue` accepts factories to delay the value instantiation

/cc @paynecrl97

### DIFF
--- a/Fluid/MemberAccessStrategyExtensions.cs
+++ b/Fluid/MemberAccessStrategyExtensions.cs
@@ -2,6 +2,7 @@
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Reflection;
+using System.Threading.Tasks;
 using Fluid.Accessors;
 
 namespace Fluid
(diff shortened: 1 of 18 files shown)


Fluid/MemberAccessStrategyExtensions.cs
```diff
@@ -162,6 +163,51 @@ public static void Register<T>(this IMemberAccessStrategy strategy, Func<T, stri
...
+        }
+
+        /// <summary>
+        /// Registers a type with a <see cref="Func{T, string, Task{Object}}"/> to retrieve any of
+        /// its property values.
+        /// </summary>
+        /// <param name="type">The type to register.</param>
+        /// <param name="name">The name of the property</param>
+        /// <param name="accessor">The <see cref="Func{T, string, Task{Object}}"/> instance used to retrieve the value.</param>
+        public static void Register<T>(this IMemberAccessStrategy strategy, string name, Func<T, string, Task<object>> accessor)
```
paynecrl97: Accessor should be `Func<T, Task<object>>` (i.e. no `string` arg). 

This argument is redundant because you're explicitly registering this accessor against a given property name, and so you already know the name of the property- no need for it to be passed in again.

**Also**- there's a copy/paste error in the summary of this method- `retrieve any of its property values` should read `retrieve the given property's value` (or similar).

See the methods I added yesterday.

Fluid/MemberAccessStrategyExtensions.cs
```diff
@@ -162,6 +163,51 @@ public static void Register<T>(this IMemberAccessStrategy strategy, Func<T, stri
...
+        }
+
+        /// <summary>
+        /// Registers a type with a <see cref="Func{T, string, TemplateContext, Task{Object}}"/> to retrieve any of
+        /// its property values.
+        /// </summary>
+        /// <param name="type">The type to register.</param>
+        /// <param name="name">The name of the property</param>
+        /// <param name="accessor">The <see cref="Func{T, string, TemplateContext, Task{Object}}"/> instance used to retrieve the value.</param>
+        public static void Register<T>(this IMemberAccessStrategy strategy, string name, Func<T, string, TemplateContext, Task<object>> accessor)
```
paynecrl97: See comments on above method

Fluid/TemplateContext.cs
```diff
@@ -130,5 +130,10 @@ public void SetValue(string name, object value)
         {
             SetValue(name, FluidValue.Create(value));
         }
+
+        public void SetValue<T>(string name, Func<T> factory)
+        {
+            SetValue(name, new FactoryValue<T>(factory));
```
paynecrl97: Should we provide an overload that passes in the `TemplateContext`?

Fluid/Values/FactoryValue.cs
```diff
@@ -0,0 +1,95 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.IO;
+using System.Text.Encodings.Web;
+using System.Threading.Tasks;
+
+namespace Fluid.Values
+{
+    public class FactoryValue<T> : FluidValue
```
paynecrl97: Should we also add a static `Create` method to maintain consistency with other `FuildValue`s?

Fluid/TemplateContext.cs
sebastienros: Not possible

Fluid/Values/FactoryValue.cs
sebastienros: Couldn't find any consistency so far, and that's ok like this,

sebastienros:
I updated the extension methods, I might have been too greedy on the generics though, but it works right now.

PR #77, opened 2018-08-05
Override the Create method to accept encode paramter

siyamandayubi:
In Orchard.Core, https://github.com/OrchardCMS/OrchardCore/blob/dev/src/OrchardCore.Modules/OrchardCore.Liquid/Startup.cs

we need to specify the encode parameter as by default, it encodes all input strings.

### DIFF
--- a/Fluid/Values/FluidValue.cs
+++ b/Fluid/Values/FluidValue.cs
@@ -52,6 +52,11 @@ namespace Fluid.Values
 
         public static FluidValue Create(object value)
         {
+            return Create(value, true);
+        }
+
+        public static FluidValue Create(object value, bool encode)
+        {
             if (value == null)
             {
                 return NilValue.Instance;
@@ -121,7 +126,7 @@ namespace Fluid.Values
 
                         case IEnumerable enumerable:
                             var fluidValues = new List<FluidValue>();
-                            foreach(var item in enumerable)
+                            foreach (var item in enumerable)
                             {
                                 fluidValues.Add(Create(item));
                             }
@@ -134,7 +139,7 @@ namespace Fluid.Values
                     return new DateTimeValue((DateTime)value);
                 case TypeCode.Char:
                 case TypeCode.String:
-                    return new StringValue(Convert.ToString(value, CultureInfo.InvariantCulture));
+                    return new StringValue(Convert.ToString(value, CultureInfo.InvariantCulture), encode);
                 default:
                     throw new InvalidOperationException();
             }


Fluid/Values/FluidValue.cs
```diff
@@ -121,7 +126,7 @@ public static FluidValue Create(object value)
 
                         case IEnumerable enumerable:
                             var fluidValues = new List<FluidValue>();
-                            foreach(var item in enumerable)
+                            foreach (var item in enumerable)
```
hishamco: Extra space after `foreach`

Fluid/Values/FluidValue.cs
```diff
@@ -51,6 +51,11 @@ public virtual bool IsNil()
         }
 
         public static FluidValue Create(object value)
+        {
+            return Create(value, true);
+        }
+
+        public static FluidValue Create(object value, bool encode)
```
hishamco: We can make it an optional parameter
```csharp
public static FluidValue Create(object value, bool encode = false)
```

siyamandayubi: Since signature Create(object value) has been used in several Lambda expressions as a type, making it to have optional parameter will force us to refactor all the cases. I think overriding the method is simpler solution and wouldn't break any usage

sebastienros:
I'd say that you will use `false` if you know it's an already encoded __string__ which means you know it's a sting and you can just return `new StringValue(value, false)`

siyamandayubi:
The use-case is in Orchard, where we use this Create method to add a TypeMapping for JObjects. We would not know the type of the property  in advance as it is used in Workflows (for example CreateContent activity). If we wouldn't override it here, (and then use the override version in Orchard) then we have two options to be able to display unicode content in Orchard. Either decode the contents before rendering, or copy the logic of the Create method to the added TypeMapping.

sebastienros:
You say uncode content, but do you mean HTML? In the case of HTML then the idea is that your template will use the `raw` tag or filter to output it. Otherwise it would make strings insecure by default.

siyamandayubi:
I mean, string with unicode characters.
For example, in the template, I have the following expression : Properties['name'] and let's say the value is something like this: سیامند ایوبی 
(btw, it is my name in persian characters)
In this case, the Fluid encodes the characters to something like #423;#533;#6443...;

And Orchard also renders it like that. 
#423;#533;#6443...;

sebastienros:
That makes more sense now. Looks like there is a double encoding then. It should encode once, not twice, there must be a bug somewhere which is not what you are fixing. This PR is just hiding the issue, I will try to repro with your example, that's good information.

sebastienros:
In Orchard did you just type this text in a text field and render it with a fluid template? Can you tell me how?

sebastienros:
I tried to repro the issue in orchard without success. I tried in a Title, in a Body (which was using `| raw` as it's supposed to contain html) then in a Text field, and every time I could see the correct text: `سیامند ایوبی` which was never encoded like you are stating. So I will need these repro steps ;)

sebastienros:
Let me add more information, in the case of the body, because I use `raw` the text is no encoded at all, and the raw unicode is rendered. However for the title and text field for which I didn't use `raw` then I could see `&#x633;&#x6CC;&#x627;&#x645;&#x646;&#x62F; &#x627;&#x6CC;&#x648;&#x628;&#x6CC;` but only in the HTML, so the final output on the browser works just fine. This is the behavior of the default html encoder in ASP.NET. I assume your issue is that you actually see `&#x633;&#x6CC;&#x627;&#x645;&#x646;&#x62F; &#x627;&#x6CC;&#x648;&#x628;&#x6CC;` in the browser output, right ? in which case I'll need a repro.

siyamandayubi:
I think these steps will reproduce it, 

- Let's  assume we have an JObject value in the Properties of a workflow. Let's say in Properties with a the key 'a'. and Let's assume the JObject has a property name with value سیامند ایوبی

- then we need a CreateContent activity in the workflow. Let's bind the title of new Content to {{ Properties['a']['name']}}

- By running the workflow, the new generated ContentItem will contain encoded characters &#x633;&#x6CC;&#x627;&#x645;&#x646

- Then by Viewing the ContentItem in Orchard, we will see &#x633;&#x6CC;&#x627;&#x645;&#x646.

- I guess the key here to repro it is using JObject.

I tried to fix it by creating ILiquidFilter that decode the value, didn't help. Then I tried to hide it :) using the generated PR.

here is the Template in the workflow

 ```
{
  "TitlePart": {
    "Title": "{{ Properties['Current_Video']['snippet']['title'] | decode }}"
  },
  "YoutubeVideo":{
  	"Youtube": {
			"EmbeddedAddress":"https://youtube.com/embed/{{ Properties['Current_Video'].id.videoId }}",
			"RawAddress":"https://youtube.com/watch?v={{ Properties['Current_Video'].id.videoId }}"
		},
	   "Description":{
			"Text":"{{ Properties['Current_Video']['snippet']['description'] }}"
		}
	},
	"ContainedPart":{
		"ListContentItemId": "{{Properties['contentItem_id']}}"
	}
}
```
And Properties['Current_Video'] contains a JObject

**Comment by siyamandayubi** [comment]
and the decode in the above example, (my failed try to avoid it), contains this

  ```
  public class DecodeFilter : ILiquidFilter
    {
        public async Task<FluidValue> ProcessAsync(FluidValue input, FilterArguments arguments, TemplateContext ctx)
        {
            string result = HttpUtility.HtmlDecode(input.ToStringValue());
            return await Task.FromResult(new StringValue(result));
        }
    }
```
The Filter doesn't work, because the value will be encoded after this

sebastienros:
Nope, the issue is in the workflow module, it's rendering some json string using liquid but is using the default encoder. The task should use the existing option while rendering the template to not html encode the result as it's not html.

siyamandayubi:
I attached the Fluid to Orchard and debug what happened in the Fluid

I thought, it was the fault of 
FluidValue.TypeMappings.Add(typeof(JValue), o => FluidValue.Create(((JValue)o).Value));

because for JObject, anyway it encodes the value, and maybe if we could say that 
FluidValue.TypeMappings.Add(typeof(JValue), o => FluidValue.Create(((JValue)o).Value, false));

then, it will not encode them.

But now I confused

PR #78, opened 2018-08-10
case multiline fix

dr0wzee:
This fixes issue in multiline case statement

{% case City %}
{%when 'Kyiv'%}...


sebastienros:
I will add a unit test first.

sebastienros:
I finally found how to reproduce your issue, fixed it and added unit tests. Thanks

PR #80, opened 2018-08-17
Fix incorrect position in parser error messages

### LINKED ISSUE #69: Incorrect position in validation messages
When parsing the following block of code:

```
<div>
	<span>
		{{{ model }}
	</span>
</div>
```

I'm getting an error: `Invalid character: '{'. at line:0, col:2`

This is a little unclear, since the error actually occurred in line 2. It looks like the parser returns character position in this particular Liquid expression, but not in the whole template.

It would be great to track the original position.

jeremycook:
Fixes #69


sebastienros:
Good job sir!!!

hishamco:
👍

PR #81, opened 2018-08-28
Use `Type.FullName` instead of `Type.Name`.

t123yh:
Consider the following situation:

```csharp
class Test<T> { class Subclass { public int Someprop { get; set; } } }
...

context.MemberAccessStrategy.Register(typeof(Test<int>.Subclass));
context.MemberAccessStrategy.Register(typeof(Test<string>.Subclass));
```

If we add accessors for `Test<int>.Subclass.Someprop`
and `Test<string>.Subclass.Someprop`, both will have name
`Subclass-Someprop`, which will lead to runtime error
`System.Reflection.TargetException: Object does not match target type.`
Using Type.FullName could address this problem.


hishamco:
Nice catch /cc @sebastienros

PR #117, opened 2019-02-24
Improving Render performance

sebastienros:
Speed is 30% better, allocations are divided by 10.

/cc @lahma


Fluid/Ast/ForStatement.cs
```diff
@@ -59,7 +63,18 @@ public override async Task<Completion> WriteToAsync(TextWriter writer, TextEncod
             {
                 int start = Convert.ToInt32((await Range.From.EvaluateAsync(context)).ToNumberValue());
                 int end = Convert.ToInt32((await Range.To.EvaluateAsync(context)).ToNumberValue());
-                elements = Enumerable.Range(start, end - start + 1).Select(x => new NumberValue(x));
+
```
lahma: Oh my more LINQ 😉 Logic would be quite much harder without it, must admit.

Fluid/Ast/ForStatement.cs
```diff
@@ -59,7 +63,18 @@ public override async Task<Completion> WriteToAsync(TextWriter writer, TextEncod
...
+                // Cache range
+                if (_rangeElements == null || _rangeStart != start || _rangeEnd != end)
+                {
+                    _rangeElements = elements = Enumerable.Range(start, end - start + 1).Select(x => new NumberValue(x));
```
lahma: Do you want to cache the initial projection template which really does nothing or the one that `ToList` produces at the end?

Fluid/Ast/ForStatement.cs
```diff
@@ -95,9 +110,10 @@ public override async Task<Completion> WriteToAsync(TextWriter writer, TextEncod
...
             {
-                var forloop = new Dictionary<string, FluidValue>();
-                forloop.Add("length", new NumberValue(list.Count));
-                context.SetValue("forloop", new DictionaryValue(new FluidValueDictionaryFluidIndexable(forloop)));
+                var forloop = new ForLoopValue();
+
+                forloop.Length = list.Count;
+                context.SetValue("forloop", forloop);
 
                 for (var i = 0; i < list.Count; i++)
```
lahma: nit: list.Count is accessed many times and it requires lookup to Count as compiler doesn't know if it's changing

Fluid/Ast/ForStatement.cs
lahma: Ah I see that Limit etc can be dynamic, this can be hard to cache then. Obvious logic to cache the list would be if all logic rules are literal or null? Don't know if that's the usual case.

Fluid/Ast/ForStatement.cs
```diff
@@ -147,5 +163,66 @@ public override async Task<Completion> WriteToAsync(TextWriter writer, TextEncod
...
+            public override string ToStringValue()
+            {
+                return "forloop";
+            }
+
+            public override ValueTask<FluidValue> GetValueAsync(string name, TemplateContext context)
+            {
+                switch (name)
+                {
+                    case "length": return new ValueTask<FluidValue>(new NumberValue(Length));
```
lahma: Would probably make sense to have `NumberValue.Create()` that would use lookups for common values between -1 and something, like in Jint. `NumberValue` seems to be immutable.

Fluid/Ast/ForStatement.cs
```diff
@@ -147,5 +163,66 @@ public override async Task<Completion> WriteToAsync(TextWriter writer, TextEncod
...
+                    case "index": return new ValueTask<FluidValue>(new NumberValue(Index));
+                    case "index0": return new ValueTask<FluidValue>(new NumberValue(Index0));
+                    case "rindex": return new ValueTask<FluidValue>(new NumberValue(RIndex));
+                    case "rindex0": return new ValueTask<FluidValue>(new NumberValue(RIndex0));
+                    case "first": return new ValueTask<FluidValue>(new BooleanValue(First));
```
lahma: `BooleanValue.Create(bool)` resolving to `BooleanValue.True` or `BooleanValue.False`?

Fluid/Ast/MemberExpression.cs
```diff
@@ -12,20 +12,22 @@ public MemberExpression(params MemberSegment[] segments)
...
 
-        public override async Task<FluidValue> EvaluateAsync(TemplateContext context)
+        public override async ValueTask<FluidValue> EvaluateAsync(TemplateContext context)
         {
             FluidValue value = null;
 
-            foreach(var segment in Segments)
+            var length = Segments.Length;
+
+            for (var i = 0; i< length; i++)
```
lahma: again foreach might be easier to get right

Fluid/DefaultFluidParser.cs
```diff
@@ -1018,7 +1033,8 @@ public static FilterArgument BuildFilterArgument(ParseTreeNode node)
         {
             string identifier = null;
             Expression term = null;
-            if (node.ChildNodes[0].Term.Name == "identifier")
+
+            if (String.Equals(node.ChildNodes[0].Term.Name, "identifier", StringComparison.Ordinal))
```
lahma: I think I once benchmarked equality and raw `==` seemed to be a solid choice, YMMV

Fluid/MemberAccessStrategy.cs
```diff
@@ -1,31 +1,22 @@
...
 using System.Collections.Generic;
 using System.Reflection;
+using System.Threading;
 
 namespace Fluid
 {
     public class MemberAccessStrategy : IMemberAccessStrategy
     {
-        private IDictionary<Type, IDictionary<string, IMemberAccessor>> _map;
+        private Dictionary<Type, Dictionary<string, IMemberAccessor>> _map;
```
lahma: Nice, direct access is always better.

Fluid/MemberAccessStrategy.cs
```diff
@@ -60,30 +59,31 @@ public IMemberAccessor GetAccessor(Type type, string name)
...
-            }
+                var localMap = _map;
 
-            typeMap[name] = getter;
-        }
+                // If it is already being used, we clone the structure
+                if (_initialized)
+                {
+                    localMap = new Dictionary<Type, Dictionary<string, IMemberAccessor>>(_map);
+                    foreach (var entry in localMap)
```
lahma: somehow it would make more sense to me to iterate _map and intialize localMap with size instead of _map, now I guess localMap construction calculates hash codes and fills by buckets just to do it again in the next loop.

Fluid/MemberAccessStrategyExtensions.cs
```diff
@@ -1,44 +1,44 @@
...
             {
-                var list = new List<string>();
+
+                list = new List<string>();
 
                 foreach (var propertyInfo in type.GetTypeInfo().GetProperties(BindingFlags.Public | BindingFlags.Instance))
                 {
                     list.Add(propertyInfo.Name);
-                    _namedAccessors.TryAdd($"({type.FullName})-{propertyInfo.Name}", new MethodInfoAccessor(propertyInfo.GetGetMethod()));
+                    _namedAccessors[$"({type.FullName})-{propertyInfo.Name}"] = new MethodInfoAccessor(propertyInfo.GetGetMethod());
```
lahma: $ is probably slower than string.Concat(1, 2, 3)..

Fluid/MemberAccessStrategyExtensions.cs
```diff
@@ -1,44 +1,44 @@
...
                 }
 
                 foreach (var fieldInfo in type.GetTypeInfo().GetFields(BindingFlags.Public | BindingFlags.Instance))
                 {
                     list.Add(fieldInfo.Name);
-                    _namedAccessors.TryAdd($"({type.FullName})-{fieldInfo.Name}", new DelegateAccessor((o, n) => fieldInfo.GetValue(o)));
+                    _namedAccessors[$"({type.FullName})-{fieldInfo.Name}"] = new DelegateAccessor((o, n) => fieldInfo.GetValue(o));
```
lahma: again $ might be slower here. Also the prefix `({type.FullName})-` is same for all members

Fluid/MemberAccessStrategyExtensions.cs
```diff
@@ -1,44 +1,44 @@
...
+
+            return list;
         }
 
         public static IMemberAccessor GetNamedAccessor(Type type, string name)
         {
-            IMemberAccessor result = null;
-
-            return _namedAccessors.GetOrAdd($"({type.FullName})-{name}", key =>
+            if (!_namedAccessors.TryGetValue($"({type.FullName})-{name}", out var result))
```
lahma: maybe benchmark the $ usage against string.Concat?

lahma: Or might value tuple be good candidate for access too? requires benchmarking I guess.

Fluid/NullEncoder.cs
```diff
@@ -1,77 +1,35 @@
...
-            {
-                return;
-            }
-
-            output.Write(value.Substring(startIndex, characterCount));
-        }
+        public override int MaxOutputCharactersPerInputCharacter => throw new NotImplementedException();
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public override unsafe int FindFirstCharacterToEncode(char* text, int textLength)
```
lahma: virtual methods cannot be inlined, so attribute does little here

Fluid/Ast/ForStatement.cs
```diff
@@ -46,7 +47,10 @@ public class ForStatement : TagStatement
         public LiteralExpression Offset { get; }
```
lahma: one thing I noticed is that many types like `LiteralExpression` have not been marked as sealed. if it were sealed, JIT could determine that here it would be direct method call to `LiteralExpression` instead of being prepared it to be a virtual call to some class defined outside of seen scope. so ideally seal when not expected to be extended outside of this library.

Fluid/NullEncoder.cs
sebastienros: https://github.com/dotnet/corefx/blob/master/src/System.Text.Encodings.Web/src/System/Text/Encodings/Web/HtmlEncoder.cs#L89

???

Fluid/MemberAccessStrategyExtensions.cs
sebastienros: It's a cached path, perf doesn't matter

Fluid/MemberAccessStrategyExtensions.cs
sebastienros: Cached path

Fluid/MemberAccessStrategyExtensions.cs
sebastienros: I think Value Type might be better, for allocations at least

Fluid/Ast/ForStatement.cs
sebastienros: Correct, and remove the default constructor to always go through a lookup?

Fluid/NullEncoder.cs
lahma: I probably need to double check with sharplab, but http://blogs.microsoft.co.il/sasha/2012/01/20/aggressive-inlining-in-the-clr-45-jit/ . Compiler intrisicts are a different thing but that example you gave is from lib, weird...

Fluid/Ast/ForStatement.cs
lahma: Create could be internal static and ensure lib only uses it, but ctr allowed for outside use and extension. Probably two versions, uint (fast array lookup) and double (slow needs to check if integral, like in Jint)

Fluid/NullEncoder.cs
sebastienros: Inlined only if sealed I have been told. I made a PR on MVC with the numbers.

lahma: OK, I guess it's safe to inline when called method is bound to compile-time known method on a sealed class then. Got to test different JIT versions... Thanks for sharing, one always learns more 🙂

lahma:
Looking good, did you get the final numbers? Did number and boolean caching help with allocations? As a side not, just noticed that many classes inherting from FluidValue haven't been sealed, might be worth to do it.

sebastienros:
Should be the same as the numbers I shared. But I will try to get scriban's results updated on their site because they were already old. I will also update my benchmarks section on the readme.

The only part where there is some improvement to make is the parser, but it's a framework (irony) and there is nothing much to do. It allocates too much (doesn't use span). But in the end it's not really useful as still faster than rendering, and in cases where it matters templates are cached.

On the rendering side it's 3x faster than scriban's on my machine, with their benchmark script. I am lagging on the parsing, but parser+render is still 2x as fast.

sebastienros:
Oh, and I broke something, it doesn't work in orchard, I need to find why :/ some infinite loop from the symptoms

xoofx:
shit, fluid is striking back! 😅 

I had to improve a bit the performance of scriban, so that it doesn't look too bad now compare to fluid 😀 
I have also updated the benchmarks, overall pretty close now, good job! But there are still a few areas that could be improved in scriban, but at some point, it will be difficult to do better without removing features.

cheers! 😊

PR #123, opened 2019-03-16
Add missing Math filters

hishamco:
According to the [Liquid docs](https://help.shopify.com/en/themes/liquid/filters/math-filters), I think we missed the top 3 math filters: `abs`, `at_least` and `at_most`. This PR is to add them to the `NumberFilters`

### DIFF
--- a/Fluid/Filters/NumberFilters.cs
+++ b/Fluid/Filters/NumberFilters.cs
@@ -24,23 +24,23 @@ namespace Fluid.Filters
 
         public static FluidValue Abs(FluidValue input, FilterArguments arguments, TemplateContext context)
         {
-            return NumberValue.Create(Math.Abs(input.ToNumberValue()), true);
+            var integral = input is NumberValue numberValue && numberValue.IsIntegral;
+
+            return NumberValue.Create(Math.Abs(input.ToNumberValue()), integral);
         }
 
         public static FluidValue AtLeast(FluidValue input, FilterArguments arguments, TemplateContext context)
         {
             var first = arguments.At(0);
-            double num = first.ToNumberValue();
 
-            return NumberValue.Create(Math.Max(input.ToNumberValue(), num), true);
+            return input.ToNumberValue() < first.ToNumberValue() ? first : input;
         }
 
         public static FluidValue AtMost(FluidValue input, FilterArguments arguments, TemplateContext context)
         {
             var first = arguments.At(0);
-            double num = first.ToNumberValue();
 
-            return NumberValue.Create(Math.Min(input.ToNumberValue(), num), true);
+            return input.ToNumberValue() > first.ToNumberValue() ? first : input;
         }
 
         public static FluidValue Ceil(FluidValue input, FilterArguments arguments, TemplateContext context)
--- a/Fluid/Filters/NumberFilters.cs
+++ b/Fluid/Filters/NumberFilters.cs
@@ -32,15 +32,7 @@ namespace Fluid.Filters
             var first = arguments.At(0);
             double num = first.ToNumberValue();
 
-            if (first is NumberValue number)
-            {
-                if (number.IsIntegral)
-                {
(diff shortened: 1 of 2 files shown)


hishamco:
@sebastienros could you please review and merge if it's fine

jeffolmstead:
@sebastienros I was just going to report these as missing filters but @hishamco already has, thanks! Do you think these will be merged soon and show up in Orchard Core? I have templates based on them but can use alternate methods if it won't be merged soon.

hishamco:
I was waiting for Sebastian since 13 days to merge, seems he is busy

Fluid/Filters/NumberFilters.cs
```diff
@@ -19,6 +22,43 @@ public static FilterCollection WithNumberFilters(this FilterCollection filters)
             return filters;
         }
 
+        public static FluidValue Abs(FluidValue input, FilterArguments arguments, TemplateContext context)
+        {
+            return NumberValue.Create(Math.Abs(input.ToNumberValue()), true);
```
sebastienros: Why the true flag for integral? Add a unit test with a decimal value to prove it's wrong

Fluid/Filters/NumberFilters.cs
```diff
@@ -19,6 +22,43 @@ public static FilterCollection WithNumberFilters(this FilterCollection filters)
...
+        }
+
+        public static FluidValue AtLeast(FluidValue input, FilterArguments arguments, TemplateContext context)
+        {
+            var first = arguments.At(0);
+            double num = first.ToNumberValue();
+
+            if (first is NumberValue number)
```
sebastienros: It doesn't need to instantiate any number. Just "compare" the two ToNumber values and return whichever argument is valid.

Fluid/Filters/NumberFilters.cs
hishamco: `Abs()` can be applied to `double` too, I just add few test cases to prove that's work

sebastienros: I was talking abou the "the true flag". The fact that the unit test worked might be a bug with this flag actually.

Fluid/Filters/NumberFilters.cs
```diff
@@ -19,6 +22,27 @@ public static FilterCollection WithNumberFilters(this FilterCollection filters)
...
+        {
+            return NumberValue.Create(Math.Abs(input.ToNumberValue()), true);
+        }
+
+        public static FluidValue AtLeast(FluidValue input, FilterArguments arguments, TemplateContext context)
+        {
+            var first = arguments.At(0);
+            double num = first.ToNumberValue();
+
+            return NumberValue.Create(Math.Max(input.ToNumberValue(), num), true);
```
sebastienros: That's not what I meant. You shouldn't have to call `Create` at all. Or use `Max` incidently.

Fluid/Filters/NumberFilters.cs
hishamco: Why the flag should be `false`?!!

Fluid/Filters/NumberFilters.cs
hishamco: Is what you mean to use `Compare` method?

Fluid/Filters/NumberFilters.cs
sebastienros: I think the best way to "keep" the integral property consistent would be to flow its value in the call. Something like this:
```
var integral = (input is NumberValue numberValue && numberValue.IsIntegral;
return NumberValue.Create(Math.Abs(numberValue.ToNumberValue()), integral);
```

Fluid/Filters/NumberFilters.cs
sebastienros:
```
var first = arguments.At(0);

return input.ToNumberValue() < first.ToNumberValue()
  ? first
  : input
  ;
```

PR #132, opened 2019-04-05
Add tag and block helpers

### LINKED ISSUE #130: Introduce helper methods for adding tags and blocks
Adding tags and blocks is tedious little bit, it would be nice if we have something like:

```csharp
services.AddMvc().AddFluid()
    .AddTags(t => t.RegisterTag<MyTag>("mytag"))
    .AddBlock(t => t.RegisterBlock<MyBlock>("myblock"));
```

/cc @sebastienros

hishamco:
Fixes #130


Fluid.MvcSample/Startup.cs
```diff
@@ -12,14 +12,14 @@ public class Startup
...
             TemplateContext.GlobalMemberAccessStrategy.Register<Person>();
-            FluidViewTemplate.Factory.RegisterTag<MyTag>("mytag");
         }
 
         // This method gets called by the runtime. Use this method to add services to the container.
         // For more information on how to configure your application, visit https://go.microsoft.com/fwlink/?LinkID=398940
         public void ConfigureServices(IServiceCollection services)
         {
-            services.AddMvc().AddFluid();
+            services.AddMvc().AddFluid()
```
sebastienros: What about have the options inside the call to AddFluid, and also only having `Add` that would accept either a tag or a block ?

hishamco:
IMHO `FluidViewEngineOptions` is something related to the fluid view, adding `WithXXX` is sort of extending the fluid with extra functionality

sebastienros:
You are right, oversight from my side

sebastienros:
But the AddTag and AddBlock are both in the Mvc part of fluid

hishamco:
Yep, something like `AddRazorPages` to `AddMvc` or `WithTemplateEngine` to `AddWebForm`. I knew your point may be because `AddFluid` return `IMvcBuilder` what if we return a new type such as `IFluideBuilder`?

sebastienros:
I don't understand why `AddFluid(o => o.Add<MyTag>())` is an issue?

hishamco:
It's not a big issue :) if you want us to go with this we should make more meaningful `AddTag`, `AddBlock` .. etc

PR #144, opened 2019-08-19
Fix generic Register not registering members

lahma:
There was still a problem with the registration. The generic version didn't do what API was promising. It was only adding the type. Now calling the non-generic version which also adds the members.


sebastienros:
Oh!!! I assume that's also why I should have more unit tests ;)

PR #150, opened 2019-09-07
Feature/money filter

jafin:
Add money filters similar to https://help.shopify.com/en/themes/liquid/filters/money-filters  (except conversion expects compatible decimal type, shopify uses integer)

Differences to above implementation, if it makes more sense to conform to spec & match to use integers instead of decimals then can do the change.

**Shopify**

```
{{ 145 | money_with_currency }}
```

**returns**
```
$1.45
```

**this PR;**
```
{{ 1.45 | money_with_currency }}
```

**returns**
```
$1.45
```

### DIFF
--- a/Fluid.Tests/Fluid.Tests.csproj
+++ b/Fluid.Tests/Fluid.Tests.csproj
@@ -7,6 +7,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.FileProviders.Abstractions" Version="1.1.1" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.8.0" />
+    <PackageReference Include="Shouldly" Version="3.0.2" />
     <PackageReference Include="xunit" Version="2.4.0" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.0" />
     <PackageReference Include="xunit.analyzers" Version="0.10.0" />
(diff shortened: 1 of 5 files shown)


hishamco:
@Skrypt seems this what you asked for few months ago

sebastienros:
There are many issues with the PR.
- the filters do not behave the same way as shopify, though their names match exactly these, so we won't be to point to their documentation
- on shopify they are supposed to follow a specific onfigurable pattern, so maybe these should be in the end system (like we do in OrchardCore for some of them)
- the formatting is done manually by finding dots or zeros, where it should use the standard C# formatting patterns

I could be happy without a way to set custom patterns, but not the other options.

jafin:
@sebastienros I'll change to use integers instead of decimal to align to the shopify documentation.

Do you know which format parameter exists in the standard formatting patterns for trimming trailing zeros?  I could not find one.  The idea submitted came from https://stackoverflow.com/questions/3104615/best-way-to-display-decimal-without-trailing-zeroes/38439847

And lastly should these even be part of this lib, given it is not in the standard liquid syntax but in the shopify variant?  I did it in response to a Issue here.

sebastienros:
I merged some dotnet filters, like `format_number` that will allow that. For instance `123 | format_number: "C"` to output `$123.00`

jafin:
Thanks @sebastienros , I'll close this PR as it's not required.

PR #151, opened 2019-09-07
feat(DateFormat): Add flag parameters to date formats  fixes #14

### LINKED ISSUE #14: Support Ruby date format flags
```
-  don't pad a numerical output.
_  use spaces for padding.
0  use zeros for padding.
^  upcase the result string.
#  change case.
:  use colons for %z.
```

https://apidock.com/ruby/DateTime/strftime

jafin:
BREAKING CHANGE %z previous included a colon in output, the docs for %z output '+0900'  , %:z will output '+09:00' (and also capital %Z)  which is the same output as %z did previously. This is to align with the docs for strftime (https://apidock.com/ruby/DateTime/strftime)

```
-  don't pad a numerical output.
_  use spaces for padding.
^  upcase the result string.
:  use colons for %z.
```
Not done (due to unable to find good examples of usage of '#', and that 0 padding appears to be a default?)
```
# change case.
0  use zeros for padding.
```

### DIFF
--- a/Fluid/Filters/MiscFilters.cs
+++ b/Fluid/Filters/MiscFilters.cs
@@ -141,7 +141,12 @@ namespace Fluid.Filters
                 sb.Builder.EnsureCapacity(format.Length * 2);
                 var result = sb.Builder;
                 var percent = false;
-    
+
+                var removeLeadingZerosFlag = false;
+                var useSpaceForPaddingFlag = false;
+                var upperCaseFlag = false;
+                var useColonsForZeeDirectiveFlag = false;
+
                 for (var i = 0; i < format.Length; i++)
                 {
                     var c = format[i];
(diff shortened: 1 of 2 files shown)


Fluid/Filters/MiscFilters.cs
```diff
@@ -160,13 +165,72 @@ public static FluidValue Date(FluidValue input, FilterArguments arguments, Templ
...
-                            case 'B': result.Append(context.CultureInfo.DateTimeFormat.MonthNames[value.Month - 1]); break;
-                            case 'c': result.Append(value.ToString("F", context.CultureInfo)); break;
+                            case '^': upperCaseFlag = true; continue;
+                            case '-': removeLeadingZerosFlag = true; continue;
+                            case '_': useSpaceForPaddingFlag = true; continue;
+                            case ':': useColonsForZeeDirectiveFlag = true; continue;
+                            case 'a':
+                            {
+                                result.Append(
+                                    upperCaseFlag
```
sebastienros: Could you rewrite your changes by reusing the computational code only once, like this:

```
var dayName = context.CultureInfo.DateTimeFormat.AbbreviatedDayNames[(int) value.DayOfWeek];
result.Append(upperCaseFlag ? dayName.ToUpper() : dayName); 
```

Or use an `if` statement if there are more than two cases.

Thanks

Issue #148, opened 2019-08-31
FluidTemplate.TryParse and FluidTemplateExtensions.Render throw some unexpected exceptions


Metalnem:
**FluidTemplate.TryParse** and **FluidTemplateExtensions.Render** can throw bunch of unexpected exceptions. The ones I've found so far are:

- **ArgumentException**
- **ArgumentOutOfRangeException**
- **DivideByZeroException**
- **NullReferenceException**
- **OverflowException**

To reproduce all of them, run the following program on .NET Core 2.2 with the latest Fluid beta (Fluid.Core 1.0.0-beta-9588):

```csharp
using System;
using System.Collections.Generic;

namespace Fluid.Fuzz
{
	public class User
	{
		public string String { get; set; }
		public int Integer { get; set; }
		public List<double> Doubles { get; set; }
	}

	public class Program
	{
		public static void Main(string[] args)
		{
			var model = new User
			{
				String = "ABC",
				Integer = 123,
				Doubles = new List<double> { 1.1, 2.2, 3.3 }
			};

			var templates = new List<string>
			{
				"{{0|remove}}",
				"{{0|modulo}}",
				"{%endfor%}",
				"{%comment%}{%",
				"<p>{{false|divided_by|modulo|urlode}}<<"
			};

			foreach (var template in templates)
			{
				Run(model, template);
			}
		}

		private static void Run(User model, string template)
		{
			try
			{
				if (FluidTemplate.TryParse(template, out var result))
				{
					TemplateContext.GlobalMemberAccessStrategy.Register<User>();
					result.Render(new TemplateContext { Model = model });
				}
			}
			catch (Exception ex)
			{
				Console.WriteLine(ex);
			}
		}
	}
}
```

It should produce output similar to this:

```
System.ArgumentException: String cannot be of zero length.
Parameter name: oldValue
   at System.String.Replace(String oldValue, String newValue)
   at Fluid.Filters.StringFilters.Remove(FluidValue input, FilterArguments arguments, TemplateContext context)
   at Fluid.FilterCollection.<>c__DisplayClass1_0.<AddFilter>b__0(FluidValue input, FilterArguments arguments, TemplateContext context)
   at Fluid.Ast.FilterExpression.EvaluateAsync(TemplateContext context)
   at Fluid.Ast.OutputStatement.WriteToAsync(TextWriter writer, TextEncoder encoder, TemplateContext context)
[...]

PR #152, opened 2019-09-09
fix (FuzzTests) Various fixes for #148 fuzz report.

jafin:
Changes to correct issues identified in #148 from the fuzz testing.
* Guard on Modulo for required arguments.
* Guard on remove for required arguments.
* GetHashCode fix on NilValue which caused stackoverflow
* Maths.Abs on StringValue GetIndex when it attempts to index negative values.
* Parse check for missing start tag on 'for'
* Throw Parser error on missing end tag
* Fuzz test cases added


sebastienros:
Thanks @jafin, I have pushed all the changes you handled here in a different commit that covers each test case individually.

sebastienros:
Feel free to file more issues if you think I forgot something: https://github.com/sebastienros/fluid/commit/4931af9b05996ef6486f94d0167120fd786fb577

Issue #153, opened 2019-09-09
Add strip_newlines filter


jptissot:
Add strip_newlines filter 

https://shopify.github.io/liquid/filters/strip_newlines/

PR #155, opened 2019-09-10
feat (strip_newlines filter): Add strip_newlines filter

jafin:
Add strip_newlines filter  (https://shopify.github.io/liquid/filters/strip_newlines/)    #153

### DIFF
--- a/Fluid.Tests/StringFiltersTests.cs
+++ b/Fluid.Tests/StringFiltersTests.cs
@@ -84,7 +84,23 @@ namespace Fluid.Tests
 
             Assert.Equal("Hello World", result.ToStringValue());
         }
-                
+
+        [Fact]
+        public void StripNewLines()
+        {
+            var input = new StringValue(@"
+Hello
+world
+");
+
+            var arguments = new FilterArguments();
+            var context = new TemplateContext();
+
+            var result = StringFilters.StripNewLines(input, arguments, context);
+
+            Assert.Equal("Helloworld", result.ToStringValue());
+        }
+
         [Fact]
         public void NewLineToBr()
         {
--- a/Fluid.Tests/StringFiltersTests.cs
+++ b/Fluid.Tests/StringFiltersTests.cs
@@ -92,7 +92,6 @@ namespace Fluid.Tests
 Hello
 world
 ");
-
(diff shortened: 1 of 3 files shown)


Fluid.Tests/TemplateTests.cs
```diff
@@ -288,6 +288,8 @@ public Task ShouldEvaluateAssignStatement(string source, string expected)
             return CheckAsync(source, expected);
         }
 
+
```
hishamco: Remove extra line

Fluid/Filters/StringFilters.cs
```diff
@@ -147,6 +149,11 @@ public static FluidValue Strip(FluidValue input, FilterArguments arguments, Temp
             return new StringValue(input.ToStringValue().Trim());
         }
 
+        public static FluidValue StripNewLines(FluidValue input, FilterArguments arguments, TemplateContext context)
+        {
+            return new StringValue(Regex.Replace(input.ToStringValue(),@"\r\n?|\n",""));
```
hishamco: Can we avoid using Regex?

Fluid.Tests/StringFiltersTests.cs
```diff
@@ -84,7 +84,23 @@ public void Strip()
...
+        public void StripNewLines()
+        {
+            var input = new StringValue(@"
+Hello
+world
+");
+
+            var arguments = new FilterArguments();
+            var context = new TemplateContext();
+
```
hishamco: Remove extra lines

Fluid/Filters/StringFilters.cs
sebastienros:
```suggestion
            var result = input.ToStringValue();
            
            if (result.Contains("\r"))
            {
                result = result.Replace("\r", "");
            }

            if (result.Contains("\n"))
            {
                result = result.Replace("\n", "");
            }

            return new StringValue(result);
```

jafin: Removed Regex replaced with above.

jafin: done replaced with search/replace.

jptissot:
fixes #153

PR #156, opened 2019-09-19
Reduce state machine creation

lahma:
Making code a bit uglier, but now detecting fast path where async targets are already resolved without need for `await`, this removes the state machine from call site. When looking at OC's traces it's a deeeep call chain because of state machinery, this PR paves way for tweaking OC behavior.

Profiled against the benchmark test case, probably could later check binary expressions etc, don't know how often they are used. This basically removes all state machines in benchmark test case's liquid template except for `ForStatement`.

* use idiomatic ExceptionHelper
* prepare some properties and helpers for better size prediction
* add netstandard 2.1 target
* upgrade support libs

## Fluid.Benchmarks.FluidBenchmarks

| **Diff**|Method|Mean|Gen 0|Gen 1|Gen 2|Allocated|
|------- |-------|-------:|-------:|-------:|-------:|-------:|
| Old |Parse|35.85 us|7.0190|-|-|28.81 KB|
| **New** |	| **37.19 us (+4%)** | **7.0190 (0%)** | **-** | **-** | **28.81 KB (0%)** |
| Old |Render|1,158.34 us|58.5938|29.2969|29.2969|224.52 KB|
| **New** |	| **946.49 us (-18%)** | **58.5938 (0%)** | **57.6172 (+97%)** | **29.2969 (0%)** | **224.52 KB (0%)** |
| Old |ParseAndRender|1,282.51 us|58.5938|29.2969|29.2969|278.01 KB|
| **New** |	| **1,069.47 us (-17%)** | **58.5938 (0%)** | **29.2969 (0%)** | **29.2969 (0%)** | **278.01 KB (0%)** |

### DIFF
--- a/Fluid/FiltersCollection.cs
+++ b/Fluid/FiltersCollection.cs
@@ -6,7 +6,18 @@ namespace Fluid
 {
     public class FilterCollection
     {
-        private Dictionary<string, AsyncFilterDelegate> _delegates = new Dictionary<string, AsyncFilterDelegate>();
+        private readonly Dictionary<string, AsyncFilterDelegate> _delegates;
+
+        public FilterCollection(int capacity = 0)
+        {
+            _delegates = new Dictionary<string, AsyncFilterDelegate>(capacity);
+        }
+
+        public int Count => _delegates.Count;
+
+#if NETSTANDARD2_1
+        public void EnsureCapacity(int capacity) => _delegates.EnsureCapacity(capacity);
+#endif
 
         public void AddFilter(string name, FilterDelegate d)
         {
--- a/Fluid/Fluid.csproj
+++ b/Fluid/Fluid.csproj
@@ -1,8 +1,9 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard1.6;netstandard2.0;net46</TargetFrameworks>
+    <TargetFrameworks>netstandard1.6;netstandard2.0;netstandard2.1;net46</TargetFrameworks>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <LangVersion>8</LangVersion>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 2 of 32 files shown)


lahma:
## Stack before

image

## Stack after

image

Fluid/FiltersCollection.cs
```diff
@@ -6,7 +6,18 @@ namespace Fluid
...
+
+        public FilterCollection(int capacity = 0)
+        {
+            _delegates = new Dictionary<string, AsyncFilterDelegate>(capacity);
+        }
+
+        public int Count => _delegates.Count;
+
+#if NETSTANDARD2_1
+        public void EnsureCapacity(int capacity) => _delegates.EnsureCapacity(capacity);
```
sebastienros: Where is it used?

sebastienros:
Scriban's approach is to have the code be non-async by default, and use code generation to generate the async counterparts. And you can see in his benchmarks that sync is way faster than async. I also moved most of the things to valuetask which improved the perf a lot. So your changes make sense.

Fluid/FiltersCollection.cs
lahma: Not yet, but profile showed Resize calls because of adding items dynamically in Orchard, this allows triggering predicted resize before adding

lahma:
Generic async makes things uniform, but in templates a few items are I/O bound, this brings back speed with cost of maintainability/readability.

Fluid/Fluid.csproj
```diff
@@ -1,8 +1,9 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard1.6;netstandard2.0;net46</TargetFrameworks>
+    <TargetFrameworks>netstandard1.6;netstandard2.0;netstandard2.1;net46</TargetFrameworks>
```
lahma: At some point might consider dropping the netstandard1.6 altogether, net46(1) should be enough for everyone

sebastienros: Yes, I actually did that two days ago on yessql.

PR #164, opened 2019-11-28
DividedBy return an integer if the divisor is integral

hishamco:


hishamco:
@sebastienros can you merge this?

Issue #135, opened 2019-05-24
Async model members are not registered with async accessors


twgraham:
I have a use case where some of the members on my template model are Tasks that should evaluate lazily. Currently, when I register my model with Fluid (`Register<Model>`), these members are not evaluated correctly.

e.g. A `Task<string>` will be evaluated as the value of `Task<>.ToString()`. I suspect this is because these members are being registered as `PropertyInfoAccessor` or `DelegateAccessor` rather than `AsyncDelegateAccessor`.

In the below scenario, the users would only be loaded if they are used inside the template when rendering.

```Csharp
public class GroupTemplateModel
{
    public string Name { get; set; }
    public Task<List<User>> Users => AsyncCallThatReturnsUserList();

    private static Task<List<User>> AsyncCallThatReturnsUserList()
    {
        // Do stuff here...
    }
}
```

*NB: I know that it would probably be wise to cache the results of these tasks for multiple access when rendering the template - In my own code I handle this using `AsyncLazy` from `Nito.AsyncEx`*

I'd be happy to make a PR. I've solved it by making small adjustments to the `MemberAccessStrategyExtensions`.

sebastienros:
If you want to try an use `AsyncDelegateAccessor` when the result is async, please do it. Same for properties I assume.

PR #182, opened 2020-04-21
Correctly register async member accessors

twgraham:
This PR addresses issue #135. When classes are registered, any property or field that is of type `Task<>` should be registered with an `AsyncDelegateAccessor` instead of a `DelegateAccessor` or `PropertyAccessor`. This could also be extended to `ValueTask<>` in future if desired.

This change uses reflection properties that are only available in netstandard2.0+, so support for netstandard1.6 was dropped. I'm sure there are other ways to achieve this support, but this seemed the most appropriate for future support while keeping the code simple.

It also relies on dynamic boxing of tasks to get the result without throwing cast exceptions, hence a dependency on `Microsoft.CSharp` was added.

### DIFF
--- a/Common.props
+++ b/Common.props
@@ -5,7 +5,7 @@
     <Authors>Sebastien Ros</Authors>
     <PackageProjectUrl>https://github.com/sebastienros/fluid</PackageProjectUrl>
     <PackageLicenseUrl>https://github.com/sebastienros/fluid/blob/master/LICENSE.txt</PackageLicenseUrl>
-    <TargetFrameworks>netstandard1.6</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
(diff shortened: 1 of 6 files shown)


Common.props
```diff
@@ -5,7 +5,7 @@
     <Authors>Sebastien Ros</Authors>
     <PackageProjectUrl>https://github.com/sebastienros/fluid</PackageProjectUrl>
     <PackageLicenseUrl>https://github.com/sebastienros/fluid/blob/master/LICENSE.txt</PackageLicenseUrl>
-    <TargetFrameworks>netstandard1.6</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
```
sebastienros, line 8: What's the reason?

twgraham: It's because of the `.IsGenericType` usage [here](https://github.com/sebastienros/fluid/pull/182/files#diff-166c73ae62ad714b7187c2df958a14c6R44). NET 4.6 has it but netstandard1.6 doesn't.

I could make an argument for updating to netstandard2.0 but I thought I'd do this because it keeps the code simple going forward.

I can change it to be 1.6 compatible if needed.

PR #184, opened 2020-04-21
Added else clause to for statement 

flew2bits:
Per https://shopify.github.io/liquid/tags/iteration/, the for statement can have an optional else clause that is executed in the case that the collection has no elements. This PR implements the else clause handling; it includes tests for both the parse and the for statement with else.


hishamco:
Please fix the conflict before the review

flew2bits:
Working on the fix now, thanks.

flew2bits:
I've fixed the merge conflicts and got rid of some warnings and other errors.

PR #187, opened 2020-05-18
changed number type to decimal, updated tests

dshabetia:
The basic operations with decimal values using double as a base type causing calculation errors.
e.g. 
- 10.4 - 10 == 0.40000000000000036  
- 23.4-20 == 3.3999999999999986

Suggested changes of the base type from **double** to **decimal** avoid such calculation errors.

### DIFF
--- a/Fluid.Tests/NumberFiltersTests.cs
+++ b/Fluid.Tests/NumberFiltersTests.cs
@@ -13,7 +13,7 @@ namespace Fluid.Tests
         [InlineData(4.2, 4.2)]
         [InlineData(-4, 4)]
         [InlineData(-4.2, 4.2)]
-        public void Abs(double value, double expectedResult)
+        public void Abs(double value, decimal expectedResult)
         {
             var input = NumberValue.Create(value);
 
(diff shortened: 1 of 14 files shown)


sebastienros:
I have been comtemplating this change a few weeks ago actually. Was wondering if that should be an option. There might be an issue big numbers that don't fit in decimal but do in double. 

(double: 1.79769313486232E+308, decimal: 79228162514264337593543950335)

DotLiquid recently changed its behavior to be decimal by default, and switch to double when the value overflow.

I am wondering also if the "integral" support might be dropped if using decimal.

sebastienros:
What's you opinion about handling overflow exceptions when parsing the decimal and then trying to parse a double. A JsValue would then have two properties, depending on which is used. Same as with the integer today.

sebastienros:
Another solution would be to have a different type of FluildValue for decimals, and handle operation across these types. By default we would create a FluidDecimal unless it overflows. Might also be a solution for integers (if decimals don't already work for this case or if it's more optimized).

PR #192, opened 2020-06-03
Support interfaces.

SebastianStehle:
Hi, 

I would like to register allowed types with their interface.

I have different implementations of an IUser interface and also mocks in unit tests.


Fluid/Values/FluidValue.cs
```diff
@@ -210,6 +210,16 @@ public static void SetTypeMapping<T>(Func<T, FluidValue> mapping)
                 return mapping;
             }
 
+            foreach (var @interface in type.GetInterfaces())
```
sebastienros, line 213: I would have expected to use the interface type here. You can see it's not even using the base types.

SebastianStehle: I don't get it.

SebastianStehle: I added this after I tried to add a workaround by adding a custom fluid value, but it did not work either.

SebastianStehle:
Is there something, I can change to get this PR through?

sebastienros:
I'll take a look soon. Please ping me here or anywhere (gitter/twitter/email) so I don't forget if you don't see progress.

sebastienros:
I updated your PR, and will comment on it to share some thoughts

Fluid/Values/FluidValue.cs
```diff
@@ -203,13 +204,46 @@ public static void SetTypeMapping<T>(Func<T, FluidValue> mapping)
...
+                    // doesn't need to do a lookup
+                    SetTypeMapping(type, mapping);
+
+                    return mapping;
+                }
+
+                currentType = currentType.GetTypeInfo().BaseType;
+            }
+
+            foreach (var interfaceType in type.GetInterfaces())
```
sebastienros, line 235: I had commented on this part of the changes because I didn't understand why you would add anything here as it was not dealing with types at all before. My concern is that it was not even dealing with base types. Then while adding it to your PR I realized that it's very perf sensitive, and that the design was that each specific needs to be mapped to another one. 

You are caching the resolved mapping, so I think you understand. The problem now is that every time we need to create a FluidValue we'll have to seek a mapping for all base types and all interfaces.

I am thinking that if you really need this feature, we might want to handle the lookup differently.
Maybe by having a custom lookup, where you pass a predicate instead of a Type mapping.
We could also have the default type mappings (object, int, DateTime) inside this dictionary instead of the subsequent switch statement. This way we can store the results even for something that doesn't have any custom mapping, instead of seeking base types and interfaces on every occurence.

Fluid/ConcurrentMemberAccessStrategy.cs
```diff
@@ -19,18 +19,36 @@ public IMemberAccessor GetAccessor(Type type, string name)
...
                         return accessor;
                     }
                 }
 
-                type = type.GetTypeInfo().BaseType;
+                currentType = currentType.GetTypeInfo().BaseType;
+            }
+
+            // Search for accessors defined on interfaces
+            foreach (var interfaceType in type.GetTypeInfo().GetInterfaces())
```
sebastienros, line 39: I refactored this part to prevent the recursion. The end result is the same but the execution is faster.

SebastianStehle: I guess I have to dig into the documentation again, the recursion was to support interfaces that are implemented by base classes as well.

Fluid/Values/FluidValue.cs
SebastianStehle: I have a IUser interface with a property that contains claims and I want to claims to be accessible as simple properties in the template. Therefore I create the mapping in my fork. This also has the advantage that my custom FluidValue is created lazily and only when needed.

I know that it is performance critical, I have just forgotten to also add null mappings to the dictionary. If we just set 

```
localTypeMappings[type] = null;
```
before we return null it should be much better. It would also make sense to have the switch statement.

Fluid/ConcurrentMemberAccessStrategy.cs
sebastienros: I think the interfaces are for the whole hierarchy. Can you check?

Fluid/Values/FluidValue.cs
sebastienros: Have you checked how we do that in Orchard? We do the exact same thing already, unless it's not the same ;)

Fluid/ConcurrentMemberAccessStrategy.cs
SebastianStehle: Yes, checked that, you are right.

SebastianStehle: https://gist.github.com/SebastianStehle/6402c465941662cd275a7bcd5da532e3

Fluid/Values/FluidValue.cs
sebastienros: For instance:

https://github.com/OrchardCMS/OrchardCore/blob/dev/src/OrchardCore.Modules/OrchardCore.Users/Liquid/UserLiquidTemplateEventHandler.cs#L23-L26

But I think you want something like this:

https://github.com/OrchardCMS/OrchardCore/blob/dev/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/RequestLiquidTemplateEventHandler.cs#L24-L69

Which shows you don't need to change SetTypeMapping as your other changes support interfaces on the Register method.

So I think we could safely remove that part of the change.

SebastianStehle: I mean this is the difference between the type of the property value and the property type. The property type is `IUser` and the instance type is `UserImpl` or so in another assembly. In the assembly where I use liquid only IUser is known, but when Liquid loops over the properties it uses the type of the property value to check for a mapping. Therefore I need this change.

sebastienros: I completely understand that. Look further in the big chunk of code I shared, don't focus on the User example. Call Register<IUser, FluidValue>((a, b) => if (b == "Name") return new String value(a.Name)))

Because this PR also supports interfaces in Register it should work whatever the implementation is. Am I wrong?

SebastianStehle: Now I get it, sorry. 

Yes, you are right, this is actually much easier than my implementation, which is coming from the attempt to find a solution before I created this PR: https://github.com/Squidex/squidex/blob/master/backend/src/Squidex.Domain.Apps.Core.Operations/Templates/Extensions/UserFluidValue.cs

But there are still use cases where a custom value might be interesting, for example when you want to override ToString() for IUser in your template. I have such a case. A json array, that I had to implement because I wanted that the JSON representation of this array to be written to the template: https://github.com/Squidex/squidex/blob/master/backend/src/Squidex.Domain.Apps.Core.Operations/Templates/Extensions/JsonArrayFluidValue.cs

But it is not an interface and this case is so special that I can live without this change.

sebastienros: I found a solution, and introduced the concept of "value converters". You can pass any lambda, and do the test you want, then return a FluidValue, an object, or null if you couldn't convert it.

PR #196, opened 2020-06-11
Completing slice implementation

sebastienros:


sebastienros:
@dshabetia 

I started to implement Shopify's test suite, and already found some limitations. Since you have improved the `slice` filter I assume you would like to know that there are other cases that should be supported.

PR #209, opened 2020-09-29
Json filter output naming policy like MemberAccessStrategy.MemberName…

Soar360:
see JsonWithMemberNameStrategy

### DIFF
--- a/Fluid/Filters/MiscFilters.cs
+++ b/Fluid/Filters/MiscFilters.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using System.Net;
 using System.Text.RegularExpressions;
 using System.Text.Json;
+using Fluid.Json;
 using Fluid.Values;
 
 namespace Fluid.Filters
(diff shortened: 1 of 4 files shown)


Fluid/Filters/MiscFilters.cs
```diff
@@ -286,7 +287,8 @@ public static FluidValue Json(FluidValue input, FilterArguments arguments, Templ
         {
             var options = new JsonSerializerOptions
             {
-                WriteIndented = arguments.At(0).ToBooleanValue()
+                WriteIndented = arguments.At(0).ToBooleanValue(),
+                PropertyNamingPolicy = new FluidJsonNamingPolicy(context)
```
sebastienros, line 291: The internal cache won't take into the account the fact that a context can have a different MemberNameStrategy though. It feels like the cache should be higher, per MemberNameStrategy  instance, and then you wouldn't need a static cache in the FluidJsonNamingPolicy. Also maybe an immutable dictionary instead of a concurrent one to prevent locking which will probably never be an issue.

Or maybe also change the naming policy contract so it can accept a name and not a member info, that would make things much easier and not require a cache.

PR #213, opened 2020-10-08
Add source link, package icon and produce snupkg

lahma:
* source link to get source integration in IDE
* snupkg to be published in NuGet
* package icon 64x64 and attached to package


Common.props
```diff
@@ -3,12 +3,19 @@
   <PropertyGroup>
     <Copyright>Sebastien Ros</Copyright>
     <Authors>Sebastien Ros</Authors>
+
```
hishamco, line 6: Extra new line

lahma: I was trying to create a bit of a grouping between logical things, I can also remove if it bothers.

PR #223, opened 2020-12-04
Add color_to_hex filter

hishamco:


Fluid/Filters/ColorFilters.cs
```diff
@@ -25,6 +26,32 @@ public static FluidValue ToRgb(FluidValue input, FilterArguments arguments, Temp
             return new StringValue($"rgb({rgbColor.R}, {rgbColor.G}, {rgbColor.B})");
         }
 
+        public static FluidValue ToHex(FluidValue input, FilterArguments arguments, TemplateContext context)
+        {
+            var color = input.ToStringValue();
+            if (!color.StartsWith("rgb(") && color.EndsWith(")") && color.Length >= 10)
```
sebastienros, line 32: I think the test is wrong

Fluid/Filters/ColorFilters.cs
```diff
@@ -25,6 +26,32 @@ public static FluidValue ToRgb(FluidValue input, FilterArguments arguments, Temp
...
+            {
+                return NilValue.Empty;
+            }
+            
+            var rgbColor = color.Substring(4, color.Length - 5).Split(',');
```
sebastienros, line 37: nit, you can use the new spread operator. Or do we need to bump the C# version in the project?

hishamco: Seems we need to bump the version

Fluid/Filters/ColorFilters.cs
hishamco: Ya it's 13 ;) I will fix it in `color_to_hsl` PR

Fluid/Filters/ColorFilters.cs
sebastienros: not worth it then

Fluid/Filters/ColorFilters.cs
sebastienros: even the && seems fishy

sebastienros: or you are missing grouping parens

hishamco: No worry, I need to refactor this method in the filter I'm working on, I will keep you updated

Thanks

PR #225, opened 2020-12-04
Add color_to_hsl filter

hishamco:


Fluid/Filters/ColorFilters.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Drawing;
+using System.Text.RegularExpressions;
```
sebastienros, line 3: No regular expressions in Fluid. I am alergic.

hishamco: I hesitate before I use it .. time to remove ...

hishamco: AFAIK Regular Expression is not good for PERF but makes the LIVE easier ;) so hope the latest changes makes you happy

Fluid/Filters/ColorFilters.cs
```diff
@@ -10,94 +11,245 @@ public static FilterCollection WithColorFilters(this FilterCollection filters)
...
+
+            public double G { get; }
+
+            public double B { get; }
+
+            public static bool TryParse(string value, out Color color)
+            {
+                if (value.StartsWith("rgb(") && value.EndsWith(")"))
+                {
+                    var rgbColor = value.Substring(4, value.IndexOf(")") - 4).Replace(" ", String.Empty).Split(',');
```
sebastienros, line 179:
```suggestion
                    var rgbColor = value.Substring(4, value.IndexOf(")") - 4).Split(new [] { ',', ' '}, StringSplitOptions.RemoveEmptyEntries);
                    
```

sebastienros: Also `value.IndexOf(")")` would return an index, whereas Substring will want a length. A solution is just to provide the first argument, and the paren will be the last char, then add `)` in the list of chars to split from.

hishamco: Ya `Substring` want a length, I used `IndexOf` to find the index and substracts the length of `rgb(` which is 4 to get the result, I  will try to use `)` in the split chars

sebastienros: Sorry, didn't see `- 4`, don't change it then

sebastienros: Or split by adding `(` and then you don't even need substring, as the elements will be `rgb`, `r`, `g`, and `b`. And there can be spaces anywhere, even at the beginning or the end.

hishamco: > Sorry, didn't see - 4, don't change it then

No problem, I changed ;)

hishamco: > Or split by adding ( and then you don't even need substring, as the elements will be rgb, r, g, and b

This will better

hishamco:
I think we are ready to merge, then I continue working other filters as well as supporting `rgba` & `hsla`

sebastienros:
I have no idea why people need it ;)

hishamco:
There are some use case on of them I can think of is storing color schema for a web application, so your theme could be using a color variations

PR #237, opened 2020-12-13
StringValue can be initialized with null

### LINKED ISSUE #234: StringValue can be initialized with null
`ToStringValue` is not supposed to be able to return `null`, however it's possible to create a `StringValue` from a filter that returns a `new StringValue(null)`.

`FluidValue.Create` correctly converts null values to `NilValue.Instance`. We should consider making the StringValue constructor private and only use `StringValue.Create` which would return `StringValue.Empty` for null.

Returning `NilValue.Instance` for `StringValue.Create` with a null value could be an option, however I think that if `StringValue` is invoked directly, we should not have something that compares to `nil` but to `empty` and `blank`.

hishamco:
Fixes #234

### DIFF
--- a/Fluid.Tests/StringValueTests.cs
+++ b/Fluid.Tests/StringValueTests.cs
@@ -1,0 +1,22 @@
+﻿using Fluid.Values;
+using Xunit;
+
+namespace Fluid.Tests
+{
+    public class StringValueTests
+    {
+        [Theory]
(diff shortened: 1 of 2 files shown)


hishamco:
Is this look good for you Seb?

hishamco:
@sebastienros can we merge this?

PR #269, opened 2021-02-20
Replace JsonSerializer with Utf8JsonWriter

Soar360:
When using JsonSerializer, the json filter will output a type even if it is not accessible in the Liquid template. Also, the serialized property name and the property name accessible in the Liquid template may not match.
So, I replaced the JsonSerializer with Utf8JsonWriter to get the same access and property naming policy as Liquid.


sebastienros:
Can you remove all the formatting changes in the switch/case? That will conflict with the fixes I made on date time formatting.

Soar360:
I will reopen a PR for resolving this conflict.

PR #270, opened 2021-02-20
Expose scope properties names

### LINKED ISSUE #185: Add ability to check available properties in Scope
Currently, you can only get or set specific values on a scope. It would be useful if `Scope`'s could share what properties were available to them. For example, being able to check all the top level properties on the current `TemplateContext` could be used to tell the user what properties are available.

In my case, I'd like to make a JSON tag or filter that could return a JSON representation of the available data, or provided expression.

I'd propose a change like the following:
```csharp
public class Scope
{
    // ... rest of class omitted for brevity

    public List<string> Properties => _properties.Keys.ToList()
}
```

This way I could see all the properties, and get values for the properties I want.

hishamco:
Fixes #185

### DIFF
--- a/Fluid/Scope.cs
+++ b/Fluid/Scope.cs
@@ -19,6 +19,8 @@ namespace Fluid
             _parent = parent;
         }
 
+        public IEnumerable<string> Properties => _properties.Keys;
+
         /// <summary>
         /// Returns the value with the specified name in the chain of scopes, or undefined
         /// if it doesn't exist.


Soar360:
It is no longer possible to get the Scope directly in the TemplateContext. We also need a way to get all the properties of the parent Scope.

hishamco:
Right, but this will introduce a hierarchy of scopes, @sebastienros your thoughts?

PR #272, opened 2021-02-22
Allow to render different models

### LINKED ISSUE #271: FluidTemplate can not render different model
The parsed template cannot render different models .

The following code will generate an exception:

``` csharp
void Main()
{
	var template = parser.Parse("Hi {{Name}}");
	var clsA = new ClassA { Name = "ClassA" };
	var clsB = new ClassB { Name = "ClassB" };
	template.Render(new TemplateContext(clsA));
	template.Render(new TemplateContext(clsB));
}

class ClassA
{
}
class ClassB
{
}
```

### InvalidCastException

```
Unable to cast object of type 'ClassB' to type 'ClassA'.
```
Is my usage incorrect?

hishamco:
Fixes #271

### DIFF
--- a/Fluid/Ast/IdentifierSegment.cs
+++ b/Fluid/Ast/IdentifierSegment.cs
@@ -1,11 +1,13 @@
-﻿using System.Threading.Tasks;
+﻿using System;
+using System.Collections.Concurrent;
+using System.Threading.Tasks;
 using Fluid.Values;
 
 namespace Fluid.Ast
 {
     public class IdentifierSegment : MemberSegment
     {
-        private IMemberAccessor _accessor;
+        private readonly ConcurrentDictionary<Type, IMemberAccessor> _accessors = new ConcurrentDictionary<Type, IMemberAccessor>();
 
         public IdentifierSegment(string identifier)
         {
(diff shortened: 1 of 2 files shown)


Fluid/Ast/IdentifierSegment.cs
```diff
@@ -1,11 +1,13 @@
...
+using System.Collections.Concurrent;
+using System.Threading.Tasks;
 using Fluid.Values;
 
 namespace Fluid.Ast
 {
     public class IdentifierSegment : MemberSegment
     {
-        private IMemberAccessor _accessor;
+        private readonly ConcurrentDictionary<Type, IMemberAccessor> _accessors = new ConcurrentDictionary<Type, IMemberAccessor>();
```
sebastienros, line 10: It's not worth it, just replace the `_accessor` cache by a local variable and don't cache it.
Or, only reuse the cache if the _accessor and another (new) _type property are the same. Most times the model sent to a template is the same, so this will work in most cases. Just be sure to "set" the _accessor after the _type if you test if _accessor is set first.

sebastienros:
Caching is hard ;) Didn't think that the same identifier could be called on different types. Thanks for identifying the issue @hishamco

Fluid/Ast/IdentifierSegment.cs
Soar360: If I use layout and cache, then I have to accommodate multiple different models in the layout.

hishamco: > It's not worth it, just replace the _accessor cache by a local variable and don't cache it.

Oops seems I did think about the lastes changes I made, it was fine .. I will update the PR soon

Fluid/Ast/IdentifierSegment.cs
```diff
@@ -36,8 +38,17 @@ public override ValueTask<FluidValue> ResolveAsync(Scope value, TemplateContext
             if (result.IsNil() && context.Model != null)
             {
                 // Check for a custom registration
-                _accessor ??= context.Options.MemberAccessStrategy.GetAccessor(context.Model.GetType(), Identifier);
-                _accessor ??= MemberAccessStrategyExtensions.GetNamedAccessor(context.Model.GetType(), Identifier, MemberNameStrategies.Default);
+                var modelType = context.Model.GetType();
+                if (modelType != _type)
```
hishamco, line 42: I checked the type first to avoid extra computation for the accessor, this way the accessor will be computed if the model type has been changed

PR #275, opened 2021-02-25
SnakeCase Strategy Error

Soar360:


Fluid/MemberNameStrategies.cs
```diff
@@ -7,7 +7,7 @@ public class MemberNameStrategies
     {
         public static readonly MemberNameStrategy Default = RenameDefault;
         public static readonly MemberNameStrategy CamelCase = RenameCamelCase;
-        public static readonly MemberNameStrategy SnakeCase = RenameCamelCase;
+        public static readonly MemberNameStrategy SnakeCase = RenameSnake;
```
hishamco, line 10: Good catch, can your rename the method to `RenameSnakeCase`

hishamco (changes requested):

Fluid/MemberNameStrategies.cs
Soar360: done~

PR #280, opened 2021-03-01
Adds startswith/endswith operators

### LINKED ISSUE #197: Add startswith/endswith operators
For discussion

I know they are not part of shopify liquid but they could be helpful.

It seems they are implemented in PowerApps and DotLiquid.

TFleury:
Closes #197

Supports both string and array values.
Also changed the extensibility sample to another operator (exclusive or).


sebastienros:
Since your comments on the PR, I more prone to seeing lower cased __operators__, as is `startswith`.
Otherwise it's perfect, thanks.

PR #283, opened 2021-03-12
Add Json filter support for Blank and Empty.

Soar360:


Soar360:
I got a "NotSupportedException: Unrecognized "FluidValue when use json fileter

hishamco:
There's a PR merged by @sebastienros, I think it might be related to this

PR #292, opened 2021-03-29
MapAction sample

sebastienros:


srid:
@sebastienros I'm trying to test this PR in my project, and potentially give feedback. Do you know how I can reference this branch on my project?

I switched to `paket` hoping it would allow me to specifying arbitrary git dependency, but unfortunately paket's git feature is quite useless for it. So I'm using Visual Studio to add a manual clone of this branch to my project. That works, however the build fails due to .NET 6.0 use in this branch:

image

Do you know what I can do to fix this?

hishamco:
> Do you know how I can reference this branch on my project?

@srid simply you can add a new remote `https://github.com/sebastienros/fluid/tree/sebros/houdini` then your will be able to use Seb's branch

hishamco:
I'm sure the changes made to use map actions ;)

sebastienros:
I will remove the net6.0 related changes, and put these in a separate branch. Also ship what's there right now as a package.

sebastienros:
@hishamco right, I tested this on MapAction to see if we could build a custom view engine for it. It works well.

srid:
> Also ship what's there right now as a package.

Awesome, so I can just wait for nuget release?

---

@hishamco I'm quite new to dotnet :-) So I don't understand what you mean by "add a new remote"?

sebastienros:
The package is here: https://www.nuget.org/packages/Fluid.ViewEngine/2.0.0-beta-1013

You will still need to **COPY**  the code from the Fluid.MapActionViewEngine in this PR since it will show how to integrate the view engine specifically for your framework. Don't try to include the project, copy the files as it and update them. Later I will provide these as some documentation section since they are specific to each target framework. But they should all follow a similar pattern.

hishamco:
> I'm quite new to dotnet :-) So I don't understand what you mean by "add a new remote"?

@srid you can do this from Git Client, it's not a dotnet thing. The following attachment shows how you can create a remote using `Fork` client

Remote

Fluid.MapActionSample/Program.cs
```diff
@@ -0,0 +1,20 @@
...
+
+// Add services to the container.
+
+builder.Services.AddFluid();
+
+var app = builder.Build();
+
+// Configure the HTTP request pipeline.
+
+app.UseRouting();
```
davidfowl, line 11: You can remove this.

PR #301, opened 2021-04-06
Fix raw statement

### LINKED ISSUE #300: Raw tag has no output
I had an [Issue in OrchardCore](https://github.com/OrchardCMS/OrchardCore/issues/9065) that directed me to open an issue here. The `{% raw %}` and `{% endraw %}` tags no longer produce any output at all, so there is no built-in way to output curly braces and the like. This has become an issue with some of our frontend templates that use JavaScript Handlebars.

While a solution is being worked on, are there any recommendations for temporary workarounds?

sebastienros:
Fixes #300


lahma:
I probably need to make state machine removal sprint at some point 😉

Issue #325, opened 2021-04-27
Add Handlebars.NET and Razor benchmarks


sebastienros:
c.f. https://github.com/scriban/scriban/tree/master/src/Scriban.Benchmarks

I am curious to see how fast are compiled templates (HB and Razor are compiled by default) compared to the current Fluid visitor perf first. The "parsing" will definitely be slower for these engines as they have to go through the compile phase as part of the parsing.

PR #327, opened 2021-04-27
Add Handlebars benchmarks

hishamco:
Addresses #325


hishamco:
Seb please try to run the benchmarks at your end, my laptop is hanging due many tabs and projects are opened ;)

sebastienros:
It can't inherit from the base one, it requires a custom template, check the link I shared in the issue. This is not Liquid, though similar. But even though I want to check if Fluid is still faster.

hishamco:
> This is not Liquid, though similar

Oops, I forgot I thought they are quite similar

PR #336, opened 2021-05-06
Hide json member names when not accessible

deanmarcussen:
When an object is serialized with the json filter inaccessible member property names where still being written.

The values were always null, but I think the names of the properties should not be available either.

This just checks if it's a `NilValue` and skips writing the property name and value in that case.

### DIFF
--- a/Fluid/Filters/MiscFilters.cs
+++ b/Fluid/Filters/MiscFilters.cs
@@ -101,7 +101,7 @@ namespace Fluid.Filters
             {
                 return arguments.At(0);
             }
-            
+
             return input;
         }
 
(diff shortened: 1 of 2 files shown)


Fluid/Filters/MiscFilters.cs
```diff
@@ -615,6 +615,11 @@ private static async ValueTask WriteJson(Utf8JsonWriter writer, FluidValue input
                             }
 
                             var fluidValue = FluidValue.Create(value, ctx.Options);
+                            if (fluidValue.IsNil())
+                            {
+                                continue;
+                            }
```
deanmarcussen, line 621: Agh! too many formatting changes.

This is the change here. Test is above

PR #342, opened 2021-05-20
IncludeStatement_ReevaluatePathOnEachWriteToAsync

suggyd:
Simple but probably inefficient fix for the include statement not re-evaluating the Path which could be based on a MemberExpression that changes during a For...In loop.  Added unit test to demonstrate the problem, but feel additional tests involving a mock IFileProvider where we can test the number of times the file is resolved is required to prove that if the Path doesn't change the template is re-used and not reloaded.  I'm using the _identifier to compare the current path against the loaded path, again this could be better as I'm duplicating the "how to make an _identifier" logic, but I didn't want to change too much at this stage.  We could also reduce the amount of work to determine the path in the repeat cases, maybe even testing whether the path is a Literal or not.  I could have also probably moved the IF to append the view extension inside the IF (template == null...) block.  I also thought it might be useful to have a small cache of loaded templates but this is probably specific to my use case.

My example liquid for this problem:

{% for part in Item.Parts %}
   {% include part.TemplateName, Item:part %}
{% endfor %}

In this example a value of the loop item provides the name of the template to load.  In the current version, all Item.Parts are rendered using the first template instead of the one they specify.

I actually created a nasty expression block that uses reflection to null the _template field on the IncludeStatement and wrapped that around the {% include %} to temporarily resolve it.

I think I'm probably abusing Liquid, but it's a really powerful use case for me where I'm dynamically create HTML from an object tree of "Parts"

### DIFF
--- a/Fluid.Tests/IncludeStatementTests.cs
+++ b/Fluid.Tests/IncludeStatementTests.cs
@@ -63,6 +63,51 @@ shape: ''";
         }
 
         [Fact]
+        public async Task IncludeSatement_ShouldLoadCorrectTemplate_IfTheMemberExpressionValueChanges()
+        {
+            var expression = new MemberExpression(new IdentifierSegment("Firstname"));
+            var sw = new StringWriter();
+
+            var fileProvider = new MockFileProvider();
+            fileProvider.Add("_First.liquid", @"{{ 'Partial Content One' }}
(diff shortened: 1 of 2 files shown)


sebastienros:
The test might be too verbose but it still works. Thanks for finding the issue and fixing it.

PR #352, opened 2021-05-27
Prevent stack overflow in MiscFilters.WriteJson

### LINKED ISSUE #351: Json Misc Filter causes stack overflow with objects that have a recursive cycle
Given a pattern similar to:

```
public static class TestObjects
        {
            public class Node
            {
                public Node NodeRef { get; set; }
            }
            public static object RecursiveReferenceObject
            {
                get
                {
                    var parent = new Node
                    {
                        Name = "Object1",
                    };
                    var child = new Node
                    {
                        Name = "Child1",
                        NodeRef = parent
                    };
                    parent.NodeRef = child;
                    return parent;
                }
            }
        }
```

Passing `TestObjects.RecursiveReferenceObject` to `MiscFilter.WriteJson()` will result in a stack overflow.

dafergu2:
Adds in a check for a cycle when writing json for an object.  Currently, a stack overflow exception is thrown.

The fix should allow multiple properties in the object graph to have the same references, it simply stops following a specific branch of the graph when a property value results in an object reference that has been traversed more than once already within the same graph path.

Fixes #351

### DIFF
--- a/Fluid/Filters/MiscFilters.cs
+++ b/Fluid/Filters/MiscFilters.cs
@@ -542,7 +542,7 @@ namespace Fluid.Filters
             return true;
         }
 
-        private static async ValueTask WriteJson(Utf8JsonWriter writer, FluidValue input, TemplateContext ctx)
+        private static async ValueTask WriteJson(Utf8JsonWriter writer, FluidValue input, TemplateContext ctx, params object[] stack)
         {
             switch (input.Type)
             {
@@ -614,6 +614,15 @@ namespace Fluid.Filters
                                 value = access.Get(obj, name, ctx);
                             }
 
+                            if (value is object)
+                            {
+                                var occurrences = stack.GroupBy(o => o).ToDictionary(o => o.Key, o => o.Count());
+                                if (occurrences.ContainsKey(value) && occurrences[value] > 1)
+                                {
+                                    value = "Cycle detected...stopping";
+                                }
+                            }
+
                             var fluidValue = FluidValue.Create(value, ctx.Options);
                             if (fluidValue.IsNil())
                             {
@@ -621,7 +630,7 @@ namespace Fluid.Filters
                             }
 
                             writer.WritePropertyName(name);
-                            await WriteJson(writer, fluidValue, ctx);
+                            await WriteJson(writer, fluidValue, ctx, stack.Concat(new [] { obj }).ToArray());
                         }
 
                         writer.WriteEndObject();
--- a/Fluid/Filters/MiscFilters.cs
+++ b/Fluid/Filters/MiscFilters.cs
@@ -628,7 +628,8 @@ namespace Fluid.Filters
 
                             writer.WritePropertyName(name);
                             stack.Add(obj);
-                            await WriteJson(writer, fluidValue, ctx, new HashSet<object>(stack));
+                            await WriteJson(writer, fluidValue, ctx, stack);
+                            stack.Remove(obj);
                         }
 
                         writer.WriteEndObject();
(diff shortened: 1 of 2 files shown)


Fluid/Filters/MiscFilters.cs
```diff
@@ -614,14 +614,23 @@ private static async ValueTask WriteJson(Utf8JsonWriter writer, FluidValue input
                                 value = access.Get(obj, name, ctx);
                             }
 
+                            if (value is object)
+                            {
+                                var occurrences = stack.GroupBy(o => o).ToDictionary(o => o.Key, o => o.Count());
+                                if (occurrences.ContainsKey(value) && occurrences[value] > 1)
```
lahma, line 620: Wouldn't this be simpler by using just `HashSet<object>` and if `stack.Add(value)` returns false it's a cycle? Initial call site just needs to pass a `new HashSet<object>()` or check if null and initialize.

Would perform a lot better unless I'm missing something.

dafergu2: I'll update the logic shortly.

dafergu2: One issue with the `HashSet<>` approach is that once you initialize it, the same reference is being used when traversing all of the edges of the object graph.  This means that if two adjacent properties have references to the same object, a cycle will be erroneously be detected.  To solve that, a new instance needs to be created as we recurse to another call of `WriteJson`.

dafergu2: Updated to use a `HashSet<>`.

Fluid/Filters/MiscFilters.cs
```diff
@@ -614,14 +614,24 @@ private static async ValueTask WriteJson(Utf8JsonWriter writer, FluidValue input
                                 value = access.Get(obj, name, ctx);
                             }
 
+                            if (value is object)
```
sebastienros, line 617: It's always true unless value is null

sebastienros (commented):
So I assume you can just remove the check object, also it's not null since it's already checked

Fluid/Filters/MiscFilters.cs
dafergu2: This was to check for null since this is the value returned from the property accessor, not the object itself.  I had originally put that there when the `.Add()` was in that block as I didn't want to add a null value to the `stack`.

Since I move the `.Add()` to after the `fluidValue.IsNil()`, removing it as you suggested makes sense.  Also, just to verify I'm understanding the flow properly, the `obj` object that gets added to the `stack` will always be a non-primitive, non-string/date/built-in, type, correct?  Otherwise, the `FluidValue` that is passed to `WriteJson` would trigger a different case statement.  Is my logic correct there?

Thanks for the review!

Fluid/Filters/MiscFilters.cs
```diff
@@ -614,14 +614,24 @@ private static async ValueTask WriteJson(Utf8JsonWriter writer, FluidValue input
...
                             var fluidValue = FluidValue.Create(value, ctx.Options);
                             if (fluidValue.IsNil())
                             {
                                 continue;
                             }
 
                             writer.WritePropertyName(name);
-                            await WriteJson(writer, fluidValue, ctx);
+                            stack.Add(obj);
+                            await WriteJson(writer, fluidValue, ctx, new HashSet<object>(stack));
```
sebastienros, line 634: I think you can pass the same `stack` instance and remove the object from it right after the call. This will save some allocations.

Fluid/Filters/MiscFilters.cs
sebastienros: All good

Fluid/Filters/MiscFilters.cs
dafergu2: Let me give that a whirl.  I just wrote another test that covers the sibling path as well.

dafergu2: @sebastienros I added that optimization, good call!  Tests pass, including the new one I added to make sure the recursion stack doesn't bleed down a sibling property path.

PR #353, opened 2021-06-09
Skip write-only property

Soar360:
NullReferenceException is thrown when the Model has a write-only property.
```
public class Test
{
    public Int32 Id { protected get;set;}
}
```


sebastienros:
Would you mind adding a simple unit test to ensure this is not regressed later on?
You can reuse one of the model classes and add such a property on it. Somewhere in this class for instance: https://github.com/sebastienros/fluid/blob/main/Fluid.Tests/MemberAccessStrategyTests.cs#L183

PR #363, opened 2021-09-09
fix IncludesFileProvider wrong setup

nkouhi:
now you can set a separate folder for includes, for example you can include from sections folder or partial folder.


sebastienros:
Will reopen if the changes didn't fix the issue.

Issue #365, opened 2021-09-23
Implement base64 filters


sebastienros:
https://github.com/Shopify/liquid/pull/1450

sebastienros:
Implementation details for "safe" versions: https://ruby-doc.org/stdlib-2.5.3/libdoc/base64/rdoc/Base64.html#method-i-urlsafe_encode64

jptissot:
Basic implementation here: 
https://github.com/StatCan/StatCan.OrchardCore/blob/master/src/Modules/StatCan.OrchardCore.DisplayHelpers/Filters/Base64Filters.cs#L8

hishamco:
We already have `UrlEncode` and `UrlDecode` filter, I will add `Base64Encode` and `Base64Decode`

sebastienros:
> We already have UrlEncode and UrlDecode filter

It is not related. `base64_url_safe_encode` uses a custom set of chars:
```
The alphabet uses ‘-’ instead of ‘+’ and ‘_’ instead of ‘/’.
```

So for these it just needs to replace `+` and `/` by `-` and `_`. It can be a simple Replace call. The goal is to generate a sort of base64 that can be used in url. The standard base64 char set contains + and / which would make it incompatible. It also has `=` but only at the end so it's not an issue.

hishamco:
I think I will be add another PR to add `base64_url_safe_encode` as another filter

hishamco:
> It also has = but only at the end so it's not an issue.

I left the = at the end because it creates an issue during decoding

PR #366, opened 2021-09-23
Add Base64 encode & decode filters

hishamco:
Addresses #365


sebastienros:
Check the issue in case you didn't see my comments

hishamco:
I already saw your comment hope to modify the PR today, it's 05:07 AM here and I didn't sleep yet ;)

PR #367, opened 2021-09-23
Add reverse filter

hishamco:


Fluid/Filters/StringFilters.cs
```diff
@@ -127,6 +128,22 @@ public static ValueTask<FluidValue> Replace(FluidValue input, FilterArguments ar
...
+        public static ValueTask<FluidValue> Reverse(FluidValue input, FilterArguments arguments, TemplateContext context)
+        {
+            var value = input.ToStringValue();
+            if (String.IsNullOrEmpty(value))
+            {
+                return StringValue.Empty;
+            }
+            else
+            {
+                var valueAsArray = value.ToCharArray();
```
sebastienros, line 140: Note that it allocates the buffer twice (array, then new string).
There is a way to do a single allocation string String.Create. It won't work for netstandard2.0 but for the other TFMs we support it will work: https://github.com/sebastienros/parlot/blob/a08477aca75ec5efceaa9948bfae466ed045944f/src/Parlot/Character.cs#L81-L85

sebastienros: But don't copy `CreateString` just include your current solution in the netstandard20 block.

hishamco: I though you will comment about this one ;)

PR #369, opened 2021-09-23
Add MD5 filter

hishamco:


Fluid/Filters/MiscFilters.cs
```diff
@@ -721,5 +721,25 @@ public static ValueTask<FluidValue> FormatString(FluidValue input, FilterArgumen
...
+        {
+            var value = input.ToStringValue();
+            if (String.IsNullOrEmpty(value))
+            {
+                return StringValue.Empty;
+            }
+
+            using (var provider = System.Security.Cryptography.MD5.Create())
+            {
+                var builder = new StringBuilder();
```
sebastienros, line 735: Initialize size with 32 and there will be only a single array allocation.
Or use `String.Create` and then it's one less allocation better.

lahma: There's also StringBuilderPool  😉

PR #371, opened 2021-09-23
Add Sha1 & Sha256 filters

hishamco:


Fluid/Filters/MiscFilters.cs
```diff
@@ -721,5 +724,45 @@ public static ValueTask<FluidValue> FormatString(FluidValue input, FilterArgumen
...
+        {
+            var value = input.ToStringValue();
+            if (String.IsNullOrEmpty(value))
+            {
+                return StringValue.Empty;
+            }
+
+            using (var provider = System.Security.Cryptography.SHA256.Create())
+            {
+                var builder = new StringBuilder(32);
```
sebastienros, line 758: 64 for a 256bits hash

Fluid/Filters/MiscFilters.cs
```diff
@@ -721,5 +724,45 @@ public static ValueTask<FluidValue> FormatString(FluidValue input, FilterArgumen
...
+        {
+            var value = input.ToStringValue();
+            if (String.IsNullOrEmpty(value))
+            {
+                return StringValue.Empty;
+            }
+
+            using (var provider = System.Security.Cryptography.SHA1.Create())
+            {
+                var builder = new StringBuilder(32);
```
sebastienros, line 738: 40 for sha1

Fluid/Filters/MiscFilters.cs
hishamco: Oops

Fluid/Filters/MiscFilters.cs
sebastienros: Why it's important is because SB will create an array internally, of a default size, and then copy the strings you append to this array, so appending doesn't allocate as long as the array as some space. Once the array is full SB will create a bigger one (new allocation) can copy the existing one. As we know the size of the final string because each hash has its own characteristics, we can tell DB how big the array it creates should be, and this way we are allocation a single array of the correct size.

PR #372, opened 2021-09-24
Add Base64 URL safe encode & decode filters

hishamco:
Addresses #365

### DIFF
--- a/Fluid.Tests/MiscFiltersTests.cs
+++ b/Fluid.Tests/MiscFiltersTests.cs
@@ -85,6 +85,32 @@ namespace Fluid.Tests
             Assert.Equal("john@liquid.com", result.ToStringValue());
         }
 
+        [Fact]
+        public async Task Base64Encode()
+        {
+            var input = new StringValue("one two three");
+
+            var arguments = new FilterArguments();
+            var context = new TemplateContext();
+
+            var result = await MiscFilters.Base64Encode(input, arguments, context);
+
+            Assert.Equal("b25lIHR3byB0aHJlZQ==", result.ToStringValue());
+        }
+
+        [Fact]
+        public async Task Base64Decode()
+        {
+            var input = new StringValue("b25lIHR3byB0aHJlZQ==");
+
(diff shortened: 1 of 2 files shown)


Fluid.Tests/MiscFiltersTests.cs
```diff
@@ -85,6 +85,66 @@ public async Task DecodeUrl()
...
+
+            // Act
+            var result = await MiscFilters.Base64UrlSafeEncode(input, arguments, context);
+
+            // Assert
+            Assert.Equal(expected, result.ToStringValue());
+        }
+
+        [Theory]
+        [InlineData("aHR0cDovL3d3dy5kb21haW4uY29t", "http://www.domain.com")]
```
sebastienros, line 132: The fact that you are using these urls for test data means that you missed a crucial point of the filters.
And it also hides a bug in the implementation of the decodesafe filter.

It's called UrlSafe because the result of the encoding can be put in a url. Not to encode urls. The idea is that when something, anything, is b64 encoded, the result can contains `+` and `/`. The issue is not the source having a `/` or `+` but the encoded result.

So a good test here is to try an encode `a<>:a?` for instance, which results in `YTw+OmE/` with base64. Don't change your implementation, and update the unit test with that and see it failing on the `Base64UrlSafeDecode` test.

And don't add urls in the test data, that would confuse the reason of the filter.

sebastienros: To be clear, use `[InlineData("a<>:a?", "YTw-OmE_")]` for the safeurlencode one and `[InlineData("YTw-OmE_", "a<>:a?")]` for the safeurldecode.

You can also add `[InlineData("a<>:a?", "YTw+OmE?")]` for the normal base64 ones.

hishamco: I will check this ASAP

hishamco: Oops, seems I forgot to replace the special chars before I decode the string

PR #374, opened 2021-09-30
Update README.md

Soar360:
add JObjectFluidIndexable to support json filter for JObject


sebastienros:
It's fixed in this PR https://github.com/sebastienros/fluid/pull/387
There was a design flaw.

Well, now that I am thinking about it, the json filter might not be fixed though, will check.

sebastienros:
All good, and shipped on NuGet. I hope it won't break anything, but `JObject` don't require any customization anymore.

Soar360:
> All good, and shipped on NuGet. I hope it won't break anything, but `JObject` don't require any customization anymore.

Thanks!

PR #379, opened 2021-10-14
fix converting double/float values

as94dodo:
I got overflow exception when i converting too small or too high double value. We should check boundaries for fixing that.


as94dodo:
I tried to use `unchecked` and got exception again

hishamco:
Could you please add a unit tests for this PR?

PR #387, opened 2021-10-30
Convert Dictionary<string, T> to FluidValue automatically

sebastienros:
This change removed the need to define custom mappings for `JObject` and other dictionary based structures.
It's also fixing the difference in converting the `Model` property and objects that are returned when accessing its properties, which is that the `Model` would not be converted to a `FluidValue` unlike its properties. With this PR the model is also converted.

### DIFF
--- a/Fluid/Values/DictionaryDictionaryFluidIndexable.cs
+++ b/Fluid/Values/DictionaryDictionaryFluidIndexable.cs
@@ -22,7 +22,11 @@ namespace Fluid.Values
             {
                 foreach (var key in _dictionary.Keys)
                 {
-                    yield return key.ToString();
+                    // Only handle string keys since this is what TryGetValue returns
+                    if (key is string)
+                    {
+                        yield return key.ToString();
+                    }
                 }
             }
         }
--- a/Fluid/Values/FluidValue.cs
+++ b/Fluid/Values/FluidValue.cs
@@ -14,6 +14,8 @@ namespace Fluid.Values
     {
         public abstract void WriteTo(TextWriter writer, TextEncoder encoder, CultureInfo cultureInfo);
 
+        public static Dictionary<Type, Type> _genericDictionaryTypeCache = new();
+
         [Conditional("DEBUG")]
         protected static void AssertWriteToParameters(TextWriter writer, TextEncoder encoder, CultureInfo cultureInfo)
         {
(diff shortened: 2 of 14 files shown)


Fluid/Values/DictionaryDictionaryFluidIndexable.cs
```diff
@@ -22,7 +22,11 @@ public IEnumerable<string> Keys
             {
                 foreach (var key in _dictionary.Keys)
                 {
-                    yield return key.ToString();
+                    // Only handle string keys since this is what TryGetValue returns
```
lahma, line 25: This class has odd `TryGetValue`, it probably should not do`Contains` + indexing, but unrelated to this PR.

Fluid/Values/FluidValue.cs
```diff
@@ -156,8 +158,14 @@ public static FluidValue Create(object value, TemplateOptions options)
                         case DateTimeOffset dateTimeOffset:
                             return new DateTimeValue(dateTimeOffset);
 
+                        case IFormattable formattable:
+                            return new StringValue(formattable.ToString(null, options.CultureInfo.NumberFormat));
```
lahma, line 162: should the parameter be `options.CultureInfo` instead of `.NumberFormat`? `FormattableString` is IFormattable and there number format sounds odd

Fluid/Values/FluidValue.cs
```diff
@@ -167,7 +175,42 @@ public static FluidValue Create(object value, TemplateOptions options)
...
+                        {
+                            if (!_genericDictionaryTypeCache.TryGetValue(typeOfValue, out genericType))
+                            {
+                                foreach (var i in typeOfValue.GetInterfaces())
+                                {
+                                    if (i.IsGenericType && i.GetGenericTypeDefinition() == typeof(IDictionary<,>) && i.GetGenericArguments()[0] == typeof(string))
+                                    {
+                                        genericType = typeof(ObjectDictionaryFluidIndexable<>).MakeGenericType(i.GetGenericArguments()[1]);
+
+                                        // Swap the previous cache with a new copy to prevent locking on reads
```
lahma, line 194: is this swap really needed as lock is only taken on cache miss? we are also locking against the field here and the field is being changed

sebastienros: I can agree with the lock instance, but not that this is an issue, just that using a dedicated lock object prevent unnecessary second check that could go null since now everything is waiting on the same object.

However I think the swap is necessary since the dictionary is static and could be accessed in read or write concurrently. So here a dictionary is always accessed either on read or on writes. If it's not swap then another thread you read it (first statement with the `TryGetValue`) while another one is adding an entry.

Fluid/Values/FluidValue.cs
sebastienros: Thanks, somehow I missed it could accept CultureInfo directly, it confused me.

Fluid/Values/DictionaryDictionaryFluidIndexable.cs
sebastienros: Not worried about this one. It's just for cases where they don't implement anything more interesting than `IDictionary`.

Issue #247, opened 2021-01-21
Is there a plan to support paginate tags?


Soar360:
https://github.com/Shopify/liquid/blob/master/performance/shopify/paginate.rb

hishamco:
Also there's a `default_pagination` may be it's useful for OC too

sebastienros:
Please go for it if you think you can implement it

hishamco:
Reopeing this, while it's useful for OC, hope to find a time to explore its features

hishamco:
After a little reading about paginate tags, there are many things should be take in mind:

1- `paginate` tag

2- integration with `for`

3- `default_pagination` filter

This needs more work, but I need to understand how it works first, seems it controls the inner for if presents

Soar360:
I have done so.
I created a new PaginationValue to handle the delayed loading of data and to generate page flip links and to do it asynchronously.
The default_pagination filter has also been implemented.
However, my solution requires the user to create a new type and derive it from the PaginationValue, which I think is not very versatile.
If you are interested in this implementation, I can post this part of the code.

hishamco:
Thanks for your initiative @Soar360 feel free to submit a PR, then we can review and see the best approach that we come up with

sebastienros:
https://shopify.dev/api/liquid/objects/paginate
https://shopify.dev/api/liquid/objects/part

PR #389, opened 2021-11-02
Implement paginate tag 

Soar360:
#247 

I implemented paginate tag and added PaginateStatement to realize this function.

Some unit tests have been added, but they may not be comprehensive enough.

The user should create a derived class for PaginateableValue and implement its abstract methods.

I wanted to make the PaginateableValue.Paginate method asynchronous, but I didn't find a good solution (I need to introduce a third-party component: Nito.AsyncEx.Context).

Maybe we can add an interface and implement a converter from this interface to PaginateableValue.

### DIFF
--- a/Fluid/Ast/PaginateStatement.cs
+++ b/Fluid/Ast/PaginateStatement.cs
@@ -1,0 +1,247 @@
+﻿using Fluid.Values;
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.IO;
+using System.Text;
+using System.Text.Encodings.Web;
+using System.Threading.Tasks;
+
+namespace Fluid.Ast
+{
+    public class PaginateStatement : Statement
+    {
+        private readonly Expression _expression;
+        private readonly long _pageSize;
+        private readonly List<Statement> _statements;
+
+        public PaginateStatement(Expression expression, long pageSize, List<Statement> statements)
+        {
+            _expression = expression ?? throw new ArgumentNullException(nameof(expression));
+            _pageSize = pageSize;
+            _statements = statements ?? new List<Statement>();
+        }
+
+        public override async ValueTask<Completion> WriteToAsync(TextWriter writer, TextEncoder encoder, TemplateContext context)
+        {
+            var value = await _expression.EvaluateAsync(context);
+            if (value == null || value is not PaginateableValue paginateableValue) return Completion.Normal;
+            context.EnterChildScope();
+            try
+            {
+                paginateableValue.PageSize = (int)_pageSize;
+
+                var data = paginateableValue.GetPaginatedData();
+                var paginate = CreatePaginate(paginateableValue, data);
+
+                context.SetValue("paginate", paginate);
+
+                await _statements.RenderStatementsAsync(writer, encoder, context);
+            }
+            finally
+            {
+                context.ReleaseScope();
+            }
+            return Completion.Normal;
+        }
+
+        private PaginateValue CreatePaginate(PaginateableValue value, PaginatedData data)
+        {
+            var ret = new PaginateValue
+            {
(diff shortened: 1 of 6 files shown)


Fluid/Filters/MiscFilters.cs
```diff
@@ -842,5 +847,77 @@ public static ValueTask<FluidValue> Sha256(FluidValue input, FilterArguments arg
...
+                    var title = paginate.Previous.Title;
+
+                    if (arguments.HasNamed("previous"))
+                    {
+                        var str = arguments["previous"].ToStringValue();
+                        if (!string.IsNullOrWhiteSpace(str)) title = str;
+                    }
+
+                    builder.AppendFormat(
+                            "<span class=\"prev\"><a href=\"{0}\">{1}</a></span>",
```
sebastienros, line 874: Any piece of text that will be rendered and could be customized should go in options. There could be a sub-object called `PaginateSettings` with many things to change.

Then it's accessible here from `context.Options.PaginateSettings`.

Fluid/Filters/MiscFilters.cs
```diff
@@ -842,5 +847,77 @@ public static ValueTask<FluidValue> Sha256(FluidValue input, FilterArguments arg
...
+                    if (arguments.HasNamed("next"))
+                    {
+                        var str = arguments["next"].ToStringValue();
+                        if (!string.IsNullOrWhiteSpace(str)) title = str;
+                    }
+
+                    builder.AppendFormat(
+                            "<span class=\"next\"><a href=\"{0}\">{1}</a></span>",
+                            HttpUtility.HtmlAttributeEncode(paginate.Previous.Url),
+                            HttpUtility.HtmlEncode(title)
```
sebastienros, line 915: The encoder should also be part of the options if not already available somewhere else.

Fluid/Ast/PaginateStatement.cs
```diff
@@ -0,0 +1,249 @@
...
+            _pageSize = pageSize;
+            _statements = statements ?? new List<Statement>();
+        }
+
+        public long PageSize => _pageSize;
+
+        public override async ValueTask<Completion> WriteToAsync(TextWriter writer, TextEncoder encoder, TemplateContext context)
+        {
+            var value = await _expression.EvaluateAsync(context);
+            if (value == null || value is not PaginateableValue paginateableValue) return Completion.Normal;
```
sebastienros, line 30: I think this should work with standard collection values (Array, Dictionary, even strings) but it should handle a custom `IPaginate` if necessary to customize the results (somehow like you did here).

sebastienros:
I couldn't find the code where the `for` tag is used and shows how the current value is replaced by the page, like in this example https://github.com/Shopify/liquid/blob/eab13a07d9861a38d993d2749ae25f06ff76426b/performance/tests/vogue/blog.liquid

Fluid/Ast/PaginateStatement.cs
Soar360: IPaginate needs to return the current page, but the standard collection does not provide this value (unless we agree on a value in the context). It also requires a method to generate a URL to populate the paginate object.

Soar360:
> I couldn't find the code where the `for` tag is used and shows how the current value is replaced by the page, like in this example https://github.com/Shopify/liquid/blob/eab13a07d9861a38d993d2749ae25f06ff76426b/performance/tests/vogue/blog.liquid

We don't need to replace it because PaginateableValue is an array by default and can be used for the for tag.

Fluid/Filters/MiscFilters.cs
Soar360: Maybe we can create a delegate to implement this function. This requires the PaginateValue and PartValue to be exposed.

PR #393, opened 2021-11-07
Fix 'reverse' naming collision

### LINKED ISSUE #392: reverse bug?
image

but:

{% assign my_array = "abc,123" | split: "," %}
{{ my_array | reverse | join: "," }}

Output:
image

hishamco:
Fixes #392

### DIFF
--- a/Fluid/Filters/ArrayFilters.cs
+++ b/Fluid/Filters/ArrayFilters.cs
@@ -15,7 +15,6 @@ namespace Fluid.Filters
             filters.AddFilter("last", Last);
             filters.AddFilter("concat", Concat);
             filters.AddFilter("map", Map);
-            filters.AddFilter("reverse", Reverse);
             filters.AddFilter("size", Size);
             filters.AddFilter("sort", Sort);
             filters.AddFilter("sort_natural", SortNatural);
@@ -130,22 +129,6 @@ namespace Fluid.Filters
             }
 
             return new ArrayValue(list);
-        }
-
-        public static ValueTask<FluidValue> Reverse(FluidValue input, FilterArguments arguments, TemplateContext context)
-        {
-            if (input.Type == FluidValues.Array)
-            {
-                return new ArrayValue(input.Enumerate().Reverse());
-            }
-            else if (input.Type == FluidValues.String)
-            {
-                return StringFilters.Reverse(input, arguments, context);
-            }
-            else
-            {
-                return input;
-            }
         }
 
         public static async ValueTask<FluidValue> Size(FluidValue input, FilterArguments arguments, TemplateContext context)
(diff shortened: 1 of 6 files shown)


hishamco:
I will move the filter into `MiscFilter` then should be ready to merge

sebastienros:
Got some new knowledge, and `reverse` doesn't work on strings as we thought. It is really an Array filter, and when applied on strings it returns an array of chard. Check this sample https://shopify.github.io/liquid/filters/reverse/ and the correct behavior in LiquidJs: https://liquidjs.com/playground.html#e3sgImZvbyIgfCByZXZlcnNlIH19,e30=

The current implementation is wrong, it will probably surprise people using Fluid currently.

hishamco:
You are right it works for Arrays only, but when I did for String I reference some Liquid docs, let me check please

hishamco:
Seb seems I miss from where I reference this, I think we have two choices here:

1. Let `reverse` applied for arrays only
2. Let `reverse` applied for arrays and strings

I think we already introduce new filters or functionality that not included in https://shopify.github.io/liquid, correct me if I'm wrong

sebastienros:
It works for strings too, it returns an Array of chars, not a reversed string.

sebastienros:
But it can be kept in the arrays filters class.

hishamco:
@sebastienros can we merge this?

Fluid/Filters/ArrayFilters.cs
```diff
@@ -103,6 +103,45 @@ public static async ValueTask<FluidValue> Map(FluidValue input, FilterArguments
...
+                    return StringValue.Empty;
+                }
+                else
+                {
+                    String reversedString;
+                    var valueAsArray = value.ToCharArray();
+#if NETSTANDARD2_0
+                    Array.Reverse(valueAsArray);
+
+                    reversedString = new String(valueAsArray);
```
sebastienros, line 126: Wrong. I mentioned that in another comment, this filter when applied on strings actually returns an array of chars, not a string.

sebastienros:
This comment points to examples: https://github.com/sebastienros/fluid/pull/393#issuecomment-962644292
`'abc' | reverse` returns `['c', 'b', 'a']`

Fluid/Filters/ArrayFilters.cs
hishamco: Oops seems I forgot to push the local changes

hishamco:
While we don't have `FluidChar` I used `FluidString` instead, we could probably introduce it - if it's necessary - while `FluidString` may used it internally

sebastienros:
@hishamco can you try to add this test?

```
{% assign my_array = "abc,123" | split: "," %}
{{ my_array | reverse | join: "," }}
```

I am not sure this PR is actually fixing the issue, though it is fixing the reverse on strings.

sebastienros:
I did and it's working

hishamco:
Seems I late to respond due my Internet connection unstable, thanks for testing that case too

PR #394, opened 2021-11-11
Json filter should skip static properties.

Soar360:
The nonparametric GetProperties() method returns the static properties together. If a custom MemberAccessStrategy is applied to an object, unexpected member names will be passed in. Therefore, JSON filters should skip these static properties.


hishamco:
Could you please provide a unit test for this?

Fluid.Tests/MiscFiltersTests.cs
```diff
@@ -754,5 +766,11 @@ private class JsonAccessStrategy
             public string Null { get; set; }
             public string Hidden { get; set; } = "Hidden";
         }
+
+        private class JsonWithStaticMember
+        {
+            public static Int32 StaticMember { get; set; } = 1;
```
hishamco, line 772: Can you use `int` instead

hishamco (approved):
LGTM after the tiny fix

PR #395, opened 2021-11-18
Upgrade NuGet packages

hishamco:


Fluid.Tests/Fluid.Tests.csproj
```diff
@@ -7,12 +7,12 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.FileProviders.Abstractions" Version="1.1.1" />
```
sebastienros, line 10: It's a low version on purpose. Unless there is a security issue, extensions should use the smallest possible version such that users don't have to bump their versions. In the end the one from the dotnet runtime will be used.

Fluid.Tests/Fluid.Tests.csproj
```diff
@@ -7,12 +7,12 @@
...
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.3.0" />
-    <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
-    <PackageReference Include="xunit" Version="2.4.1" />
```
sebastienros, line 13: No need to use beta releases, unless it's fixing an issue we are having.

PR #398, opened 2021-11-18
Remove TimeZoneConverter dependency on net6.0

sebastienros:


Fluid/Filters/MiscFilters.cs
```diff
@@ -264,10 +265,27 @@ public static ValueTask<FluidValue> ChangeTimeZone(FluidValue input, FilterArgum
 
             var timeZone = arguments.At(0).ToStringValue();
 
-            if (!TZConvert.TryGetTimeZoneInfo(timeZone, out var timeZoneInfo)) return new DateTimeValue(value);
+
+#if HAS_TIMEZONE_API
```
lahma, line 269: What about the new NET_6_OR_GREATER kind of directives?

hishamco: It would be much better

sebastienros: Good and bad at the same time. When I created it I thought there was a new API to do it, but actually it's just a new behavior of the existing API. So yes in this case `NET_6_OR_GREATER` will be better. However in general it's better to use custom values like this.

sebastienros:
Can someone help me understand why the test is failing on windows for GH? It works locally for me, on Windows 11. The call to `FindSystemTimeZoneById` throws an exception for `"America/Los_Angeles"` but it's a valid IANA name.

hishamco:
I will try to test it in my Windows

hishamco:
Same here, still investigating on this ...

hishamco:
I tried `Pacific Standard Time` the exception is gone, but still some unit tests are failing

sebastienros:
`Pacific Standard Time` is a standard windows TZ name. The tests here try to use the IANA names, which is supported on Windows only since dotnet 6. It's like the test is not actually using dotnet 6 :/

sebastienros:
https://devblogs.microsoft.com/dotnet/cross-platform-time-zones-with-net-core/

lahma:
Might be worthwhile to check how it behaves on newer windows image, like 2022 preview: https://github.blog/changelog/2021-08-23-github-actions-windows-server-2022-with-visual-studio-2022-is-now-available-on-github-hosted-runners-public-beta/ . The `windows-latest` is Windows Server 2019.

sebastienros:
Might be related to https://github.com/dotnet/runtime/issues/60175

sebastienros:
The logs seem to indicate Windows 10 October 2018 Update (Microsoft Windows 10.0.17763).
But anyhow since the behavior will depend on some OS settings, I won't be able to make it work. I can only force ICU when referencing the package like this:

```
<PackageReference Include="Microsoft.ICU.ICU4C.Runtime" Version="68.2.0.9" /><RuntimeHostConfigurationOption Include="System.Globalization.AppLocalIcu" Value="68.2" />
```

And I am not even sure that's something that can be done from a package. So for now we'll have to keep the dependency on TimeZoneConverter.

PR #401, opened 2021-11-21
Fix file change detection

### LINKED ISSUE #400: Feedback for MinimalViews
I tested https://github.com/sebastienros/fluid/blob/main/MinimalApis.LiquidViews/README.md

Suggestions: 
- All liquid files are meant to be in a Views folder? Please say it at the beginning.
- Move `From /Views/_ViewStart.liquid` outside the code block (or else we would also paste it).
- Update assets (Bootstrap beta 3) to obtain a nice layout or may be a ClassLess css like https://github.com/mblode/marx if BS is too cumbersome (Preview some classless css : https://dohliam.github.io/dropin-minimal-css/)

I had to rebuild to actually see the changes made in the Liquid files. Have I missed anything?
Is Hot reload supposed to work?

What about making a dotnet new Template nuget package?

sebastienros:
Fixes #400


dodyg:
Nice. It works beautifully.

sebastienros:
Thanks for testing it

PR #402, opened 2021-11-22
Add render tag

### LINKED ISSUE #235: {% include %} is deprecated and replaced by {% render %}
https://shopify.dev/docs/themes/liquid/reference/tags/deprecated-tags
https://github.com/Shopify/liquid/blob/5532df880fda804c838a35324ba7b8ea6cf0e34f/test/integration/tags/render_tag_test.rb

Might also require the {% section %} tag

sebastienros:
Fixes #235


sebastienros:
Also completes the `include` tag by adding `with/as` and `for/as` support.
The only difference between `include` and `render` is that `render` only has access to the global scope (options.Scope) while `include` can also access the parent template's variables.

hishamco:
Why the `ForLoopStatement` got replace by `ForLoopValue`?

sebastienros:
`ForSatement` is still here, I just extracted `ForLoopValue` to its own file since I needed it in include/render tags.

PR #409, opened 2021-12-01
Support single line liquid tags

sebastienros:
[DO NOT MERGE]

This is just to track the feature which I find useful, to be able to used inlined liquid tags instead of requiring a new line before each tag.

Right now Liquid 5.1.0 doesn't support this feature.

```
{% liquid if name == 'tobi' 
    echo 'hello tobi'
    endif %}
```

Could then be written 

```
{% liquid if name == 'tobi' echo 'hello tobi' endif %}
```


sebastienros:
https://github.com/Shopify/liquid/issues/1493

PR #414, opened 2021-12-01
Fix render tag local context

sebastienros:


Fluid/Ast/RenderStatement.cs
```diff
@@ -0,0 +1,195 @@
...
+                        await _template.RenderAsync(writer, encoder, context);
+                    }
+                    finally
+                    {
+                        context.LocalScope = previousScope;
+                    }
+                }
+                else if (AssignStatements != null)
+                {
+                    foreach (var assignStatement in AssignStatements)
```
lahma, line 98: Faster to use plain `for` loop against `IList`

PR #426, opened 2021-12-11
handle/handleize filter should return a Liquid-consistent value

### LINKED ISSUE #424: handle/handleize filter doesn't return a Liquid-consistent value
The example on [this page](https://shopify.dev/api/liquid/filters/string-filters#handle-handleize) of the Liquid doc claims this --

```
{{ '100% M & Ms!!!' | handle }}
```

-- should output this:

```
100-m-ms
```

In Fluid v2.2.8, it does not. It outputs:

```
100% m & ms!!!
```

The Fluid implementation seems to be missing:

1. Removal of punctuation
2. Replacement of spaces with dashes

hishamco:
Fixes #424

/cc @deanebarker


Fluid.Tests/MiscFiltersTests.cs
```diff
@@ -527,7 +527,8 @@ public async Task DateIsRenderedWithCulture()
         [InlineData("ALLCAPS", "allcaps")]
         [InlineData("One1Two2Three3", "one1-two2-three3")]
         [InlineData("ONE1TWO2THREE3", "one1two2three3")]
-        [InlineData("First_Second_ThirdHi", "first_second_third-hi")]
+        [InlineData("First_Second_ThirdHi", "first-second-third-hi")]
```
hishamco, line 530: > The string is converted to lowercase, and any spaces and special characters replaced by hyphens (-).

According to the liquid docs, if the underscore is a special character that means it should be replaced by hypen

Fluid/Filters/MiscFilters.cs
```diff
@@ -100,7 +101,32 @@ public static FilterCollection WithMiscFilters(this FilterCollection filters)
...
+            var value = new StringBuilder(kebabCase);
+            var previousChar = value[0];
+
+            for (int i = 1; i < value.Length; i++)
+            {
+                var currentChar = value[i];
+                
+                if (Char.IsWhiteSpace(currentChar) || Char.IsPunctuation(currentChar))
+                {
+                    value.Replace(currentChar, KebabCaseSeparator);
```
sebastienros, line 114: This is doing a replace on the whole string for every occurrence of an unsupported char?

deanebarker: I didn't write that. @hishamco ?

hishamco: Seems I didn't check the whole code @deanbarker, I updated the PR

hishamco: I will reduce the number of removal by one for **n** consequence similar chars

sebastienros:
Have you looked at how the SlugService works in Orchard?

hishamco:
> Have you looked at how the SlugService works in Orchard?

Nope, but I will have a quick look now ...

hishamco:
It quite similar except I used a `StringBuilder` and `SlugService` removes the dashes and trim using `string`. I will try to rid off the `Trim` extension and trying to mimimize unnecessary iterations as much as I can

hishamco:
Finally I did it with a single loop, hope it goes much better

Fluid/Filters/MiscFilters.cs
```diff
@@ -100,7 +101,37 @@ public static FilterCollection WithMiscFilters(this FilterCollection filters)
...
+                {
+                    if (appendDash)
+                    {
+                        appendDash = false;
+                        result.Append(KebabCaseSeparator);
+                    }
+                }
+            }
+
+            var lastIndex = result.Length - 1;
```
hishamco, line 128: It could be replaced by `TrimEnd`

Fluid/Filters/MiscFilters.cs
```diff
@@ -100,7 +101,37 @@ public static FilterCollection WithMiscFilters(this FilterCollection filters)
...
+            if (result[lastIndex] == KebabCaseSeparator)
+            {
+                result.Remove(lastIndex, 1);
```
hishamco, line 131: It could be replaced by `TrimEnd`

Fluid/Filters/MiscFilters.cs
```diff
@@ -100,7 +101,37 @@ public static FilterCollection WithMiscFilters(this FilterCollection filters)
         /// </summary>
         public static ValueTask<FluidValue> Handleize(FluidValue input, FilterArguments arguments, TemplateContext context)
         {
-            return new StringValue(HtmlCaseRegex.Replace(input.ToStringValue(), "-$1$2").ToLowerInvariant());
+            var value = HtmlCaseRegex
```
sebastienros, line 104: Is this regex required?

hishamco: Yep, this is what you did before to convert into Kebab-Case, but we could remove after this PR. For now it's fine to keep it, because it reduces a lot of code ;)

One comment about that `RegEx`, can we set its options into compiled?

sebastienros: Pretty sure it could be removed with what you have written.

hishamco:
@sebastienros is there anything else prevent from merge this?

PR #429, opened 2021-12-14
Fix async pattern in RenderAsync

### LINKED ISSUE #428: FluidTemplateExtensions.RenderAsync premature release scope
The method:
```csharp
public static ValueTask<string> RenderAsync(this IFluidTemplate template, TemplateContext context, TextEncoder encoder)
```
ReleaseScope in finally block.

When my template running too long, scope premature released .

I write some test code :

```csharp

parser.RegisterEmptyBlock("sleep", async (statements, writer, encoder, context) =>
 {
	 context.EnterChildScope();
	 context.IncrementSteps();
	 context.SetValue("id", "123");
	 await Task.Delay(100);
	 await statements.RenderStatementsAsync(writer, encoder, context);
	 context.ReleaseScope();
	 return Completion.Normal;
 });


var context = new TemplateContext {};
context.SetValue("id","000");
var template = parser.Parse(@"{{id}}
{%sleep%}{{id}}{%endsleep%}
{{id}}
");
 
var output = await template.RenderAsync(context);

Console.WriteLine(output);
```

Output: 
```
000
000
000
```

But,the right output should be:

```
000
123
000
```

sebastienros:
Fixes #428

### DIFF
--- a/Fluid/FluidTemplateExtensions.cs
+++ b/Fluid/FluidTemplateExtensions.cs
@@ -32,7 +32,7 @@ namespace Fluid
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static ValueTask<string> RenderAsync(this IFluidTemplate template, TemplateContext context, TextEncoder encoder)
+        public static async ValueTask<string> RenderAsync(this IFluidTemplate template, TemplateContext context, TextEncoder encoder)
         {
             if (context == null)
             {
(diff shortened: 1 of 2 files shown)


lahma (approved):
Looks good, pretty much the same where to I reduced the code to get it working.

Fluid/FluidTemplateExtensions.cs
```diff
@@ -44,45 +44,25 @@ public static ValueTask<string> RenderAsync(this IFluidTemplate template, Templa
...
-            }
-
             var sb = StringBuilderPool.GetInstance();
             var writer = new StringWriter(sb.Builder);
 
+            // A template is evaluated in a child scope such that the provided TemplateContext is immutable
+            context.EnterChildScope();
+
             try
             {
```
hishamco, line 54: What will be returns if an exception is occured?!!

sebastienros: And Exception

hishamco: So this intentionally?!! coz isn't throw before

hishamco: Probably we need a remark on that method to make clear to everyone

lahma: I don't think the behavior has changed here. New code just awaits to prevent control flow from getting out with unfinished continuations that could run after finally has been processed.

hishamco: What confuses me the old behavior returns the actual string from the ` StringBuilderPool` instace in both case (Exception or Without Exception) but now the string is returned if no exception occurs. Am I right?

lahma: The old code would have done the `await` on `Awaited` method which would have bubbled the exception. So both versions will throw, this one obeys template context life time.

sebastienros: Wrong, if there had been an exception, it would have been thrown by the task that is returned.
No exception is caught here, so everything will bubble up.

PR #431, opened 2021-12-17
Expose TryGetDateTimeInput to be usable by custom filters

Matthewsre:
I have a few custom filters in my solution for manipulating DateTime values and this would have been much easier if the TryGetDateTimeInput was made public or exposed.

I noticed there was another extension in this file and thought this would be a clean and simple approach that exposes this for custom filters in the Fluid.Filters namespace.

### DIFF
--- a/Fluid/Filters/MiscFilters.cs
+++ b/Fluid/Filters/MiscFilters.cs
@@ -264,7 +264,7 @@ namespace Fluid.Filters
 
         public static ValueTask<FluidValue> ChangeTimeZone(FluidValue input, FilterArguments arguments, TemplateContext context)
         {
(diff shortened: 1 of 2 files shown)


hishamco:
I agree, it will be much cleaner if we move it to `Fluid.Extensions` folder and the namespace is `Fluid.Values`

Matthewsre:
I created an Extensions folder in the Fluid project and tried adding a FluidValueExtensions class with the Fluid.Values namespace, but it was conflicting with an existing FluidValueExtensions class in the Values folder. I decided to move it into the existing FluidValueExtensions class. 

Does this work for you, or would you prefer both the new and existing extension be moved into a new Extensions folder?

PR #432, opened 2022-01-01
Fix lock contention in FluidValue.Create

lahma:
While doing performance checkup in Orchard (load test profiling agency theme) I noticed that there was quite big lock contention problem in Fluid, it originates from https://github.com/sebastienros/fluid/pull/387 where the logic never assigns anything on cache miss thus making the locking and looping happen every time for types like `ContentItem` (non-dictionary type). 

image

Fixed issue by using `Interlocked.CompareExchange`  (no locks needed) and also storing null as result for given type.  The dictionary was public, don't know why (Orchard doesn't seem to need it), made it private.

## Main

|         Method |     Mean |   Error |  StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------------- |---------:|--------:|--------:|-------:|------:|------:|----------:|
| CreateFromList | 287.9 ns | 1.22 ns | 1.08 ns | 0.0110 |     - |     - |     184 B |

## This PR

|         Method |     Mean |    Error |   StdDev |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|--------------- |---------:|---------:|---------:|-------:|------:|------:|----------:|
| CreateFromList | 46.98 ns | 0.979 ns | 1.048 ns | 0.0029 |     - |     - |      48 B |

### DIFF
--- a/Fluid.Benchmarks/FluidValueBenchmarks.cs
+++ b/Fluid.Benchmarks/FluidValueBenchmarks.cs
@@ -1,0 +1,21 @@
+using System.Collections.Generic;
+using BenchmarkDotNet.Attributes;
+using Fluid.Values;
+
+namespace Fluid.Benchmarks
+{
+    [MemoryDiagnoser]
+    public class FluidValueBenchmarks
+    {
+        // many interfaces, none of which is dictionary
+        private static readonly List<string> nonDictionaryType = new();
+
+        private static readonly TemplateOptions options = new();
+
+        [Benchmark]
+        public FluidValue CreateFromList()
+        {
+            return FluidValue.Create(nonDictionaryType, options);
+        }
+    }
+}
--- a/Fluid/Ast/LiquidStatement.cs
+++ b/Fluid/Ast/LiquidStatement.cs
@@ -15,8 +15,9 @@ namespace Fluid.Ast
         {
             context.IncrementSteps();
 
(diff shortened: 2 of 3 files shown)


sebastienros:
There is a risk with the swap technique that multiple threads can never get to an agreement with all types in the dictionary. Example with two types, `A` and `B`.
Thread 1 checks for `A`, Thread 2 checks for `B`, none is there, T1 adds `A` replaces dictionary, T2 adds `B`, replaces dictionary, `A` is not there anymore, and this could go on. This is theory, but my previous lock was preventing this. Thread 2 would be blocked until Thread 1 was done.

So ideally the only thing to fix would be the `null` in the dictionary. In terms of perf no lock would be hit once the cache is warm.

lahma:
https://docs.microsoft.com/en-us/dotnet/api/system.threading.interlocked.compareexchange?view=net-6.0#System_Threading_Interlocked_CompareExchange_System_Object__System_Object_System_Object_ - exchange only happens if field has same reference as in start, if not, someone did successful update and the current race loser will try again. But can also be changed back to double-checked locking.

sebastienros:
I understand now, thanks for the explanation. There might be other places I could use that pattern then, I have used the "switch" technique in some places knowing the limitation I was explaining. This technique solves this quite nicely.

PR #434, opened 2022-01-04
Implement functions

### LINKED ISSUE #433: Implement functions support
Adding function invocation support by accepting `()` segments with arguments (anonymous at first) such that members can be invoked. Something like `articles.take(1)` will call `InvokeAsync()` on the `FluidValue` returned by `articles.take`. This is the same pattern as indexers.

- The parser will fail if the option is not allowed to retain Liquid standard by default.
- Create a `FunctionValue` that takes a `Func<TemplateContext, FluidValue>` and whose `InvokeAsync` invokes it.
- Create a `macro` tag that can define local functions https://mozilla.github.io/nunjucks/templating.html#macro

sebastienros:
Fixes #433 

Missing:
- [ ] Macro tag to define local functions (snippets)
- [ ] Documentation

### DIFF
--- a/Fluid/Values/FunctionValue.cs
+++ b/Fluid/Values/FunctionValue.cs
@@ -1,0 +1,77 @@
+﻿using System;
+using System.Globalization;
+using System.IO;
+using System.Text.Encodings.Web;
+using System.Threading.Tasks;
+
+namespace Fluid.Values
+{
+    public sealed class FunctionValue : FluidValue
+    {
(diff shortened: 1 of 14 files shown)


Fluid/Values/FunctionValue.cs
```diff
@@ -0,0 +1,77 @@
...
+        }
+
+        public override object ToObjectValue()
+        {
+            return "";
+        }
+
+        public override string ToStringValue()
+        {
+            return "";
```
Soar360, line 53: Why not use string.Empty ?

sebastienros: It's exactly the same. Faster to type.

PR #435, opened 2022-01-05
Upgrade Parlot 0.0.19

hishamco:


sebastienros:
Currently done in the functions PR, and that's the reason for this new version. Duplicate.

PR #448, opened 2022-01-08
Make Enumerate asynchronous

### LINKED ISSUE #446: Remove obsolete methods

### LINKED ISSUE #447: Make Enumerate async
Useful for `FunctionValue`. This might not be trivial if other non-async paths already use `Enumerate` like in expressions.

sebastienros:
Fixes #447
Fixes #446

I believe some iterations can now use the Awaited pattern.

### DIFF
--- a/Fluid.Tests/ArrayFiltersTests.cs
+++ b/Fluid.Tests/ArrayFiltersTests.cs
@@ -106,7 +106,7 @@ namespace Fluid.Tests
 
             var result = ArrayFilters.Concat(input, arguments, context);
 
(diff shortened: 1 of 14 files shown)


Soar360:
Does the Contains method also need to be asynchronous?

sebastienros:
You might be right, since `contains` is called on collections, and we could see a `FunctionValue` be invoked with `contains` so it would have to be enumerated. Can you file an issue, and find some examples? I will look into it too.

Soar360:
Sure . #453

PR #450, opened 2022-01-09
Display source of error in message

### LINKED ISSUE #449: Include the template line that failed to parse in the error output
I get a lot of parse exceptions here:

https://github.com/sebastienros/fluid/blob/main/Fluid/FluidParserExtensions.cs#L22

Something like

>End of tag '}}' was expected at (2:8):

The way I have my system architected, it can be really hard to figure out _which template_ threw the error. My implementation makes multiple calls to `Template.Render`, and this error message gives little/no context as to which template caused the problem.

In the above example: line 2, position 8 _of which template?_

The most valuable thing I did was change this line of code to:

```
throw new ParseException($"{parlotError.Message} at {parlotError.Position}: {template}");
```

This outputs the full template string that threw the error. However, this is a hack, since the entire template is usually too long to put in an error message.

But I wonder if there would be some value in including the single line that threw the parse error?

sebastienros:
Fixes #449

### DIFF
--- a/Fluid.Tests/ErrorMessagesTests.cs
+++ b/Fluid.Tests/ErrorMessagesTests.cs
@@ -27,5 +27,17 @@ namespace Fluid.Tests
             Assert.False(_parser.TryParse(source, out var _, out var error));
             Assert.StartsWith(expected, error); // e.g., message then 'at (1:15)'
         }
+
+        [Theory]
+        [InlineData("  {% assign a 'b' %}")]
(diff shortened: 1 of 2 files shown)


sebastienros:
Result:

```
'=' was expected after the identifier of 'assign' at (2:14)
Source:
  {% assign a 'b' %}
```

sebastienros:
@deanebarker works for you?

deanebarker:
@sebastienros  Yes, it appears to work perfectly. Thank you.

PR #459, opened 2022-01-12
Fix keywords conflicts

sebastienros:


Fluid/FluidParser.cs
```diff
@@ -116,14 +112,23 @@ public FluidParser(FluidParserOptions parserOptions)
...
-                .Or(False.Then<Expression>(x => new LiteralExpression(BooleanValue.False)))
-                .Or(Empty.Then<Expression>(x => new LiteralExpression(EmptyValue.Instance)))
-                .Or(Blank.Then<Expression>(x => new LiteralExpression(BlankValue.Instance)))
-                .Or(Member.Then<Expression>(x => x))
+                .Or(Member.Then<Expression>(static x => {
+                    if (x.Segments.Count == 1)
+                    {
+                        switch ((x.Segments[0] as IdentifierSegment).Identifier)
+                        {
+                            case "empty": return new LiteralExpression(EmptyValue.Instance);
```
lahma, line 123: would it make sense to have `internal static` field, `return LiteralExpression.EmptyValue`?

sebastienros: Yes, these are immutable.

PR #464, opened 2022-01-13
Add rendering event

### LINKED ISSUE #455: Ability to set additional context values when using Fluid.MvcViewEngine
When using Fluid.MvcViewEngine and the new FunctionValue options. We want to add additional items to the FluidContext in an MVC context. Current there is no way to do this, outside of forking / modifying the full Fluid.MvcViewEngine project.

Consider adding a way to modify or switch the FluidRendering class - currently registered as as Singleton, specifically allowing us to define our own RenderAsync method.

This also applied to setting AmbientValues in the FluidContext too.

Thinking of  something like the below as an example....
```
public async Task RenderAsync(ViewContext viewContext, string path)
        {
            var context = new TemplateContext(_options.TemplateOptions);
            context.SetValue("ViewData", viewContext.ViewData);
            context.SetValue("ModelState", viewContext.ModelState);
            context.SetValue("Model", viewContext.ViewData.Model);
            
            var lowercase = new FunctionValue((args, context) =>
            {
                var firstArg = args.At(0).ToStringValue();
                var lower = firstArg.ToLower();
                return new ValueTask<FluidValue>(new StringValue(lower));
            });
            
            context.SetValue("LowerCase", lowercase);

            await _fluidViewRenderer.RenderViewAsync(viewContext.Writer, path, context);
        }
```

sebastienros:
Fixes #455

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,23 +23,18 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    - name: Setup .NET 2.1
+
+    - name: Setup .NET
       uses: actions/setup-dotnet@v1
       with:
-        dotnet-version: 2.1.*
-    - name: Setup .NET 3.1
-      uses: actions/setup-dotnet@v1
-      with:
-        dotnet-version: 3.1.*
-    - name: Setup .NET 5.0
-      uses: actions/setup-dotnet@v1
-      with:
-        dotnet-version: 5.0.*
-    - name: Setup .NET 6.0
-      uses: actions/setup-dotnet@v1
-      with:
-        dotnet-version: 6.0.*
+        dotnet-version: | 
+          2.1.x
+          3.1.x
+          5.0.x
+          6.0.x
+          
     - name: Test
       run: dotnet test --configuration Release
+    
     - name: Test - Compiled
       run: dotnet test --configuration Release /p:Compiled=true
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,7 +28,6 @@ jobs:
       uses: actions/setup-dotnet@v1
       with:
         dotnet-version: | 
-          2.1.x
           3.1.x
           5.0.x
           6.0.x
--- a/Fluid.MvcViewEngine/FluidMvcViewOptions.cs
+++ b/Fluid.MvcViewEngine/FluidMvcViewOptions.cs
@@ -1,10 +1,16 @@
 ﻿using Fluid.ViewEngine;
-using System.Collections.Generic;
+using Microsoft.AspNetCore.Mvc.Rendering;
+using System.Threading.Tasks;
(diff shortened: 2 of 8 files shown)


.github/workflows/build.yml
```diff
@@ -23,23 +23,18 @@ jobs:
...
-      with:
-        dotnet-version: 5.0.*
-    - name: Setup .NET 6.0
+
+    - name: Setup .NET
       uses: actions/setup-dotnet@v1
       with:
-        dotnet-version: 6.0.*
+        dotnet-version: | 
+          2.1.x
```
lahma, line 31: I think only `netcoreapp3.1;net5.0;net6.0` are referenced in code base

Fluid.MvcViewEngine/FluidMvcViewOptions.cs
```diff
@@ -1,9 +1,16 @@
...
 {
     public class FluidMvcViewOptions : FluidViewEngineOptions
     {
-        // Placeholder for future options specific to the MVC view engine
+        public delegate ValueTask RenderingMvcViewDelegate(string path, ViewContext viewContext, TemplateContext context);
+
+        /// <summary>
+        /// Gets or sets the delegate to execute when a view is rendered.
+        /// </summary>
+        public new RenderingMvcViewDelegate RenderingViewAsync { get; set; }
```
lahma, line 14: the `new` sounds like trouble, should it be separate property?

sebastienros: What kind of trouble do you see? There is no point in exposing the base one in this case since the delegate has more arguments. But I don't want it to behave randomly. I expect this property to just replace the base one. Or that the base one is null when invoked as the base type.

sebastienros: Checked the documentation and the based property will remain null, unless access directly by casting. I think that's fine in this context.

lahma: Using `new` to hide members is usually a bit of code smell as behavior bound to type used to access the thing, that is all.

markeverard:
Hey Seb, it would be really useful for us for this change to be merged into main and released via Nuget.

Fluid.ViewEngine/FluidViewEngineOptions.cs
```diff
@@ -62,5 +64,12 @@ public class FluidViewEngineOptions
         /// Gets or sets whether files should be reloaded automatically when changed. Default is <code>true</code>;
         /// </summary>
         public bool TrackFileChanges { get; set; } = true;
+
+        public delegate ValueTask RenderingViewDelegate(string path, TemplateContext context);
```
hishamco, line 68: Add missing docs

hishamco (dismissed):

PR #485, opened 2022-04-01
Fix ElseIfStatement double-evaluation in awaited context

lahma:
This code in awaited context check caused `elseIf.WriteToAsync` to always start writing, even when it should be only started if condition is true:

```c#
if (!elseIfConditionTask.IsCompletedSuccessfully)
{
    var writeTask = elseIf.WriteToAsync(writer, encoder, context);
    return AwaitedElseBranch(elseIfConditionTask, writeTask, writer, encoder, context, i + 1);
```

Parametrized test case to test async and non-async paths the get some red tests and then fixed. Now passing the `elseIf` and write task is retrieved, if needed.

We found out this when running on Linux against sync code, which was quite curious, on Windows it always worked.

### DIFF
--- a/Fluid.Tests/IfStatementTests.cs
+++ b/Fluid.Tests/IfStatementTests.cs
@@ -4,26 +4,30 @@ using System.Text.Encodings.Web;
 using System.Threading.Tasks;
 using Fluid.Ast;
 using Fluid.Values;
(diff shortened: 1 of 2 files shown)


lahma:
@sebastienros this bug was somewhat nasty and unpredictable (template writing both elsif and else content), should there be a new release?

Issue #470, opened 2022-02-01
Allow Template selection to be context aware (ASP.NET DisplayMode)


SpeedyStuart:
**Problem:**
I need to be able to use different liquid templates based on runtime information (cookie, user-agent, etc.).  Similar to `DisplayMode` in ASP.NET.

Whilst it is possible to provide a custom File Provider which contains this logic and returns appropriate templates, the Template Cache in `FluidViewRenderer` just caches templates based on path - so renders my logic useless.

**Possible change**
- Add a new delegate to `FluidViewEngineOptions` that takes the template path and returns a Template Cache key (string)
- Use this delegate (if present) in `GetFluidTemplateAsync` to generate the key for each template in the Template Cache

e.g.
```
// Allow templates to be cached by external factors
// Typically the rule would match any rules in custom IFileProvider implementations
// e.g. Serve different templates based on Authentication status / User Agent etc.
string cacheKey = path;
if (_fluidViewEngineOptions.TemplateCacheKeyProvider!=null)
{
    cacheKey = _fluidViewEngineOptions.TemplateCacheKeyProvider.Invoke(path);
}

if (cache.TemplateCache.TryGetValue(cacheKey, out var template))
{
    return template;
}

template = await ParseLiquidFileAsync(path, fileProvider, includeViewStarts);

cache.TemplateCache[cacheKey] = template;

return template;
```

SpeedyStuart:
Ah, ok.

The Mvc one.

SpeedyStuart:
Created pull request #486 for this

PR #486, opened 2022-04-04
Ability to set template cache key

SpeedyStuart:
This allows implementers to provide a delegate for generating the key that's used for template caching/retrieval.  Therefore it allows templates to be cached based on something other than (just) their path + name.
Used with a custom file provider it allows different templates to be served based on arbitrary request properties - e.g. user-agent, cookie etc.
Possible solution for #470

### DIFF
--- a/Fluid.ViewEngine/FluidViewEngineOptions.cs
+++ b/Fluid.ViewEngine/FluidViewEngineOptions.cs
@@ -74,5 +74,15 @@ namespace Fluid.ViewEngine
         /// Gets or sets the delegate to execute when a view is rendered.
         /// </summary>
         public RenderingViewDelegate RenderingViewAsync { get; set; }
+
+        /// <summary>
+        /// Represents the method that will generate a cache key for a supplied template path.  Allows templates to be cached based on any external factor.
+        /// </summary>
+        public delegate string TemplateCacheKeyProviderDelegate(string path);
+
+        /// <summary>
+        /// Gets or sets the delegate to execute when a template cache key is required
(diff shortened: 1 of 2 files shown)


hishamco:
@SpeedyStuart, please fix the conflict

Fluid.ViewEngine/FluidViewRenderer.cs
```diff
@@ -15,7 +15,6 @@ namespace Fluid.ViewEngine
     public class FluidViewRenderer : IFluidViewRenderer
     {
         private static readonly char[] PathSeparators = { '/', '\\' };
-
```
hishamco, line 18: Why?

Fluid.ViewEngine/FluidViewEngineOptions.cs
```diff
@@ -74,5 +74,15 @@ public class FluidViewEngineOptions
         /// Gets or sets the delegate to execute when a view is rendered.
         /// </summary>
         public RenderingViewDelegate RenderingViewAsync { get; set; }
+
+        /// <summary>
+        /// Represents the method that will generate a cache key for a supplied template path.  Allows templates to be cached based on any external factor.
+        /// </summary>
+        public delegate string TemplateCacheKeyProviderDelegate(string path);
```
hishamco, line 81: I'm not sure if it's a good idea to have a delegate, I'm thinking of providing another API for caching or using `IMemoryCache` where the cache key could be provided by the user

hishamco (changes requested):

PR #497, opened 2022-06-03
Make color filter independent of CultureInfo.CurrentCulture

JustusGreiberORGADATA:
I actually wanted to fix a different bug regarding layout file resolution, but all color tests were failing on my german Windows machine so this had to come first. 

Color filters need to generate and parse valid CSS values. The syntax of
these is always independent of the currently set culture, so the filters
must reflect that by using CultureInfo.InvariantCulture.

The Visual Studio auto formatter also changed stuff like 
```
if ((uint) red > 255)
```
to
```
if ((uint)red > 255)
```

There seems to be no `.editorconfig` in this repository so I don't know if this is right 🤷‍♂️. If you don't want this, I will change it back :).

### DIFF
--- a/Fluid.Tests/ColorFiltersTests.cs
+++ b/Fluid.Tests/ColorFiltersTests.cs
@@ -1,7 +1,8 @@
-﻿using Fluid.Values;
-using Fluid.Filters;
-using Xunit;
+﻿using Fluid.Filters;
+using Fluid.Values;
+using System.Globalization;
 using System.Linq;
+using Xunit;
 
 namespace Fluid.Tests
 {
@@ -34,6 +35,27 @@ namespace Fluid.Tests
 
             // Assert
             Assert.Equal(expected, result.Result.ToStringValue());
+        }
+
+        [Theory]
+        [InlineData("en-US")]
+        [InlineData("de")]
+        [InlineData("fr-FR")]
+        [InlineData("zh-Hans")]
+        public void ToRgbShouldNotBeAffectedByCurrentCulture(string culture)
+        {
+            // Arrange
+            CultureInfo.CurrentCulture = CultureInfo.GetCultureInfo(culture);
+            CultureInfo.CurrentUICulture = CultureInfo.GetCultureInfo(culture);
+
+            var input = new StringValue("hsla(0.5, 77.3%, 49.1%, 0.5)");
+            var context = new TemplateContext();
+
+            // Act
+            var result = ColorFilters.ToRgb(input, FilterArguments.Empty, context);
+
+            // Assert
+            Assert.Equal("rgba(222, 30, 28, 0.5)", result.Result.ToStringValue());
         }
 
         [Theory]
@@ -70,6 +92,27 @@ namespace Fluid.Tests
         }
 
         [Theory]
+        [InlineData("en-US")]
+        [InlineData("de")]
+        [InlineData("fr-FR")]
+        [InlineData("zh-Hans")]
+        public void ToHexShouldNotBeAffectedByCurrentCulture(string culture)
+        {
+            // Arrange
+            CultureInfo.CurrentCulture = CultureInfo.GetCultureInfo(culture);
+            CultureInfo.CurrentUICulture = CultureInfo.GetCultureInfo(culture);
+
+            var input = new StringValue("hsla(0.5, 77.3%, 49.1%, 0.5)");
+            var context = new TemplateContext();
+
+            // Act
+            var result = ColorFilters.ToHex(input, FilterArguments.Empty, context);
+
+            // Assert
+            Assert.Equal("#de1e1c", result.Result.ToStringValue());
+        }
+
+        [Theory]
         [InlineData("#fff", "hsl(0, 0%, 100%)")]
         [InlineData("#000", "hsl(0, 0%, 0%)")]
         [InlineData("#f00", "hsl(0, 100%, 50%)")]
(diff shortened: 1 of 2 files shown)


hishamco:
> I actually wanted to fix a different bug regarding layout file resolution, but all color tests were failing on my german Windows machine so this had to come first.

Could you please write down a failing test, so I could try from my side, before we dig into your changes - which supposed to fix the culture issue -

> There seems to be no .editorconfig in this repository so I don't know if this is right 🤷‍♂️. If you don't want this, I will change it back :).

I suggest to revert the changes, we could consider this in `.editorconfig` if we do so

JustusGreiberORGADATA:
> Could you please write down a failing test, so I could try from my side, before we dig into your changes - which supposed to fix the culture issue -

ok I removed the changes and replaced them with a commit that just adds a few failing tests, that should show the problem 😊.

hishamco:
@JustusGreiberORGADATA I expect you will comment your fix, seems not ;) add your fix into another commit .. time to review

Thanks

JustusGreiberORGADATA:
Added the fix again 👍

hishamco:
I surprised when I saw another file has been changes, then I realize that you did that ;) Just few minutes to finalize my review

Fluid.Tests/ColorFiltersTests.cs
```diff
@@ -36,6 +37,27 @@ public void ToRgb(string color, string expected)
...
 
+        [Theory]
+        [InlineData("en-US")]
+        [InlineData("de")]
+        [InlineData("fr-FR")]
+        [InlineData("zh-Hans")]
+        public void ToRgbShouldNotBeAffectedByCurrentCulture(string culture)
+        {
+            // Arrange
+            CultureInfo.CurrentCulture = CultureInfo.GetCultureInfo(culture);
```
hishamco, line 48: Create private method named `SetCurrentCulture()` instead of repeating the code over and over

Fluid/Filters/ColorFilters.cs
```diff
@@ -685,21 +686,29 @@ public static bool TryParse(string value, out RgbColor color)
                 if ((value.StartsWith("rgb(") || value.StartsWith("rgba(")) && value.EndsWith(")"))
                 {
                     var rgbColor = value.Split(_colorSeparators, StringSplitOptions.RemoveEmptyEntries);
+
+                    // We will parse these values with NumberStyles.Float, because it gets "close enough" to the actual CSS Standard.
+                    // In reality, we would need to also parse number like '.2' to 0.2.
```
hishamco, line 691: Remove this comment and file a new issue with this enhancement

Fluid/Filters/ColorFilters.cs
```diff
@@ -802,9 +811,19 @@ private static double QqhToRgb(double q1, double q2, double hue)
                     );
             }
 
-            public override string ToString() => A == DefaultTransperency
```
hishamco, line 805:
```csharp
public override string ToString() => A == DefaultTransperency
                ? $"rgb({R}, {G}, {B})".ToString(CultureInfo.InvariantCulture)
                : $"rgba({R}, {G}, {B}, {Math.Round(A, 1)})".ToString(CultureInfo.InvariantCulture);
```

Fluid/Filters/ColorFilters.cs
```diff
@@ -858,21 +877,32 @@ public static bool TryParse(string value, out HslColor color)
                 if ((value.StartsWith("hsl(") || value.StartsWith("hsla(")) && value.EndsWith(")"))
                 {
                     var hslColor = value.Split(_colorSeparators, StringSplitOptions.RemoveEmptyEntries);
+
+                    // We will parse these values with NumberStyles.Float, because it gets "close enough" to the actual CSS Standard.
+                    // In reality, we would need to also parse number like '.2' to 0.2.
```
hishamco, line 882: Same as above

Fluid/Filters/ColorFilters.cs
```diff
@@ -944,9 +974,20 @@ public static bool TryParse(string value, out HslColor color)
                 return new HslColor(Convert.ToInt32(h), Math.Round(s, 2), Math.Round(l, 2), rgbColor.A);
             }
 
-            public override string ToString() => A == DefaultTransparency
```
hishamco, line 947: Same as above

hishamco (changes requested):

Fluid/Filters/ColorFilters.cs
JustusGreiberORGADATA: This does not seem to work 🤔 If I add the ToString directly to the literal, the compiler seems to infer that it is a regular string instead...

hishamco (approved):
LGTM

Fluid/Filters/ColorFilters.cs
hishamco: Oh really

hishamco: If it doesn't compile which I don't  why ;) keep your previous changes

hishamco (changes requested):

Fluid/Filters/ColorFilters.cs
JustusGreiberORGADATA: Apparently there is [FormattableString.Invariant](https://docs.microsoft.com/en-us/dotnet/api/system.formattablestring.invariant) to do what you want to do. Check it out I pushed an updated version :).

Fluid/Filters/ColorFilters.cs
```diff
@@ -802,9 +804,9 @@ private static double QqhToRgb(double q1, double q2, double hue)
                     );
             }
 
-            public override string ToString() => A == DefaultTransperency
-                ? $"rgb({R}, {G}, {B})"
-                : $"rgba({R}, {G}, {B}, {Math.Round(A, 1)})";
+            public override string ToString() => (A == DefaultTransperency)
```
hishamco, line 807: No need for paranthesis in `A == DefaultTransperency`

Fluid/Filters/ColorFilters.cs
```diff
@@ -944,9 +947,9 @@ public static bool TryParse(string value, out HslColor color)
                 return new HslColor(Convert.ToInt32(h), Math.Round(s, 2), Math.Round(l, 2), rgbColor.A);
             }
 
-            public override string ToString() => A == DefaultTransparency
-                ? $"hsl({H}, {S * 100.0}%, {L * 100.0}%)"
-                : $"hsla({H}, {S * 100.0}%, {L * 100.0}%, {Math.Round(A, 1)})";
+            public override string ToString() => (A == DefaultTransparency)
```
hishamco, line 950: Same here

hishamco (changes requested):

hishamco (approved):

PR #503, opened 2022-06-29
Add translate filter and supporting infrastructure.

zxjon22:
Hi there, I had a requirement for something like See https://shopify.dev/themes/architecture/locales/storefront-locale-files to use translations from resx files.

Not sure if it's something you'd like in the base library but here's a PR for it.

- Adds a `translate` (or `t`) filter to load and use translations from an external source.
- Adds an `IResourcesProvider` to `TemplateContext` and `TemplateOptions` for the filter to load translations.
- Adds a concrete `ResxResourcesProvider` to load translations from standard .NET resx files.

There is also a `NullResourcesProvider` which simply returns the translation key unmodified - this is installed by default. I'm not 100% sure about this. I added it so that using `t` wouldn't cause a crash if a suitable `IResourcesProvider` wasnt configured but maybe it's better to throw an exception.

Use it like this:
```charp
var rp = new ResxResourcesProvider("MyApp.Resources", GetType().Assembly);
var context = new TemplateContext { ResourcesProvider = rp } ;
var result = template.Render(context);
```

```resx
<?xml version="1.0" encoding="utf-8"?>
<!-- MyApp.Resources.resx -->
<root>
    <xsd:schema id="root" xmlns="" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:msdata="urn:schemas-microsoft-com:xml-msdata">
    </xsd:schema>
    <data name="Welcome" xml:space="preserve">
        <value>Hello there!</value>
    </data>
    <data name="Footer" xml:space="preserve">
        <value>Goodbye, {0}.</value>
    </data>    
</root>
```

```liquid
<p>{{ "Welcome" | t }}</p>
<p>{{ "Goodbye" | t: "Bob"}}</p>
```

#### Result
```html
<p>Hello there!</p>
<p>Goodbye Bob.</p>
```

### DIFF
--- a/Fluid.Tests/Fluid.Tests.csproj
+++ b/Fluid.Tests/Fluid.Tests.csproj
@@ -25,4 +25,10 @@
     <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
   </ItemGroup>
 
+  <ItemGroup>
+    <EmbeddedResource Update="SR.resx">
+      <Generator></Generator>
+    </EmbeddedResource>
+  </ItemGroup>
+
 </Project>
--- a/Fluid.Tests/MiscFiltersTests.cs
+++ b/Fluid.Tests/MiscFiltersTests.cs
@@ -738,6 +738,36 @@ namespace Fluid.Tests
             Assert.Equal("c7ac4687585ab5d3d5030db5a5cfc959fdf4e608cc396f1f615db345e35adb9e", result.ToStringValue());
         }
 
+        /// <summary>
+        /// <see cref="MiscFilters.Translate"/> filter tests.
+        /// </summary>
+        /// <remarks>
+        /// The default <see cref="NullResourcesProvider"/> is used for these tests. This provider just returns the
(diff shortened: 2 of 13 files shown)


hishamco:
At the first glance I'm not sure if @sebastienros agree to use RESX for localization string

Soar360:
I think this matter should be given to the theme engine to do it, not a template engine.

hishamco:
IMHO this filter might go into OC

sebastienros:
I like what you are doing here but I don't think it should be part of the engine natively. I suggest you keep it in your source code or create a separate package that adds this support to Fluid.

For instance here is how we did it in Orchard Core, by using the ASP.NET `ViewLocalizer` https://github.com/OrchardCMS/OrchardCore/blob/main/src/OrchardCore/OrchardCore.DisplayManagement.Liquid/Filters/LiquidViewFilters.cs#L13

sebastienros:
But I might change my mind later on. I see you haven't added any external dependencies which I like.

PR #507, opened 2022-08-29
Don't add ellipsis when size is greater than number of words

### LINKED ISSUE #506: truncatewords - truncate indicator is always added, even if shorter
If i use `truncatewords` in one of my OC liquid templates, the 3 dots are always visible, even if the text is much shorter.
image

This happend to me, when i used it like this `truncatewords: 40`, but the 3 dots are also shown on a text with 20 words or something like that.
image

Changing the indicator to a custom one also doesn't change that behavior.

hishamco:
Fixes #506

### DIFF
--- a/Fluid.Tests/StringFiltersTests.cs
+++ b/Fluid.Tests/StringFiltersTests.cs
@@ -304,6 +304,7 @@ world
         [InlineData("测试测试测试测试", 5, "测试测试测试测试")]
         [InlineData("one  two\tthree\nfour", 3, "one two three...")]
         [InlineData("one two three four", 2, "one two...")]
+        [InlineData("  one two three four", 2, "one two...")]
         [InlineData("one two three four", 0, "one...")]
         public void TruncateWords(string input, object size, string output)
         {
--- a/Fluid.Tests/StringFiltersTests.cs
+++ b/Fluid.Tests/StringFiltersTests.cs
@@ -297,17 +297,20 @@ world
         }
 
         [Theory]
-        [InlineData("The cat came back the very next day", 4, "The cat came back...")]
-        [InlineData("The cat came back the very next day", 1, "The...")]
-        [InlineData("The cat came back the very next day", 0, "...")]
-        [InlineData("The cat came back the very next day", 8, "The cat came back the very next day...")]
-        [InlineData("The cat came back the very next day", 12, "The cat came back the very next day")]
-        [InlineData("The    cat came  back", 10, "The    cat came  back")]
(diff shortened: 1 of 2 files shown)


Fluid.Tests/StringFiltersTests.cs
```diff
@@ -300,7 +300,9 @@ public void TruncateWithCustomEllipsis(string input, int size, string ellipsis,
         [InlineData("The cat came back the very next day", 4, "The cat came back...")]
         [InlineData("The cat came back the very next day", 1, "The...")]
         [InlineData("The cat came back the very next day", 0, "...")]
-        [InlineData("The    cat came  back", 10, "The    cat came  back...")]
+        [InlineData("The cat came back the very next day", 8, "The cat came back the very next day...")]
```
sebastienros, line 303: Can you confirm this is the behavior in Shopify?

Fluid/Filters/StringFilters.cs
```diff
@@ -299,7 +299,10 @@ public static ValueTask<FluidValue> TruncateWords(FluidValue input, FilterArgume
                 source = "";
             }
 
-            source += ellipsis;
+            if (size <= words)
```
sebastienros, line 302: Shouldn't this be done also in the truncate filter?

Fluid.Tests/StringFiltersTests.cs
hishamco: I just tried them in Liquid JS

Fluid/Filters/StringFilters.cs
hishamco: Nope

Fluid.Tests/StringFiltersTests.cs
sebastienros: I know. LiquidJs is not bullet proof, I have files issues there when I found bugs.

Tests to add:
https://github.com/Shopify/liquid/blob/ff1c35b986451cca14d9da9b2ef772cbc9c65b24/test/integration/standard_filter_test.rb#L217-L234

And one that got removed I think and which was wrong, is to add multiple spaces, they should be reduced to a single single one, like a `String.Join(" ", String.Split(RemoveEmptyEntries))`

sebastienros: I see the test wasn't removed, but the result it wrong.

hishamco: I see, I will check the ruby code tomorrow, coz I feel sleepy ;)

PR #515, opened 2022-10-18
Add replace_last and remove_last filters

### LINKED ISSUE #494: Implement replace_last and replace_first filters
https://github.com/Shopify/liquid/pull/1422

danielmpetrov:
Bonus - add a missing assert statement in `RemovesReturnsInputWhenArgumentIsEmpty` unit test.

Resolve #494

### DIFF
--- a/Fluid.Tests/ColorFiltersTests.cs
+++ b/Fluid.Tests/ColorFiltersTests.cs
@@ -1,7 +1,7 @@
 ﻿using Fluid.Filters;
+using Fluid.Tests.Extensions;
 using Fluid.Values;
 using System.Globalization;
-using System.Linq;
 using Xunit;
 
 namespace Fluid.Tests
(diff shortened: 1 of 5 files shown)


sebastienros:
Awesome!

Can you also add the missing test cases from here:
https://github.com/Shopify/liquid/pull/1422/files#diff-783fde0f471731937b1ede37eff3107dae698e98dc76dda994391f49ca73958aR542-R566

This contains some with types that are not strings (I see it's implemented but not tested).

danielmpetrov:
Thanks for the quick feedback.

No problem, I can add those. Need to run now, but will update the PR later today or likely tomorrow.

hishamco:
Can you use `Theory` instead to examine different use cases for both unit tests

hishamco:
Can you use `Theory` instead to examine different use cases for both unit tests

sebastienros:
I added an extra unit test

Issue #517, opened 2022-10-19
Nondeterministic unit tests


danielmpetrov:
Unit tests are assuming the user has not modified the default culture settings. I have the regional format set as `English (United States)` on my machine, but have customized some of the default formats under the `Date & time` menu in the Windows settings app. This is my config:

![windows-region](https://user-images.githubusercontent.com/15656341/196750947-f9648234-0013-4fe4-b682-ad57b91f85d9.PNG)

And with these settings, five unit tests are failing for me. One example below:

![failing-test](https://user-images.githubusercontent.com/15656341/196751539-7e230782-a0d1-466a-a2ed-49b20c36cfa4.PNG)

sebastienros:
Nice find, these might be bugs or unit tests that need to force the culture (in the options, like we do for several ones already).

danielmpetrov:
The culture is specified, it's just that the expected outcome is hardcoded and doesn't match with the computer configuration.

Is this what you mean by forcing the culture - https://github.com/sebastienros/fluid/blob/main/Fluid.Tests/MiscFiltersTests.cs#L259 (this is one of the tests that fails), or am I misunderstanding?

sebastienros:
> the expected outcome is hardcoded

Hard coded as in "this is exactly what should be rendered, based on the current culture options and any server configuration.

> Is this what you mean by forcing the culture

Yes, some culture-dependent tests/templates will have that to ensure that the "hard-coded" expected result is deterministic. I am never setting the current thread culture to a weird one it might blead to other tests that are executed concurrently (there is a risk with async that the code runs on two different threads). Another option would be to run the whole set of tests with a different culture to be certain that we didnt' miss the option in some places, or that the configure culture is correctly taken into account.

danielmpetrov:
Specifying a second parameter `useUserOverride: false` fixes this issue. Acceptable solution?

```csharp
new CultureInfo("en-US", useUserOverride: false)
```

sebastienros:
Oh, I see what you mean now, that's a different issue. I am well aware of this option as we are currently discussing its usage in ASP.NET request localization and Orchard Core ;)

For the tests yes we should do that such that the environment doesn't break them.

sebastienros:
> You have added this recently, any idea why this is failing locally for me?

What culture and TFM are you running on?
Try to change to English or use a different TFM (if one works) and debug to see the differences.

hishamco:
I might handle this because I already worked in exact usecase in Orchard Core

[...]

PR #518, opened 2022-10-19
Fix local settings breaking unit tests

danielmpetrov:


danielmpetrov:
Partially resolves #517

PR #519, opened 2022-10-20
Force UTC when converting from TimeSpan

### LINKED ISSUE #517: Nondeterministic unit tests

danielmpetrov:
Fix #517

### DIFF
--- a/Fluid.Tests/MiscFiltersTests.cs
+++ b/Fluid.Tests/MiscFiltersTests.cs
@@ -410,18 +410,19 @@ namespace Fluid.Tests
         }
 
         [Theory]
-        [InlineData("0", "00:00:00.000")]
-        [InlineData("1:2", "01:02:00.000")]
-        [InlineData("1:2:3.1", "01:02:03.100")]
-        public async Task DateTimeSpan(string timespan, string expected)
+        [InlineData("0", "%H:%M:%S.%L", "00:00:00.000")]
+        [InlineData("1", "%z", "+0000")]
+        [InlineData("1:2", "%H:%M:%S.%L", "01:02:00.000")]
+        [InlineData("1:2:3.1", "%H:%M:%S.%L", "01:02:03.100")]
+        public async Task DateTimeSpan(string timespan, string format, string expected)
         {
-            // Converting to Unix time should not vary by TimeSone
+            // Converting to Unix time should not vary by TimeZone
 
             var input = FluidValue.Create(TimeSpan.Parse(timespan), new TemplateOptions());
-            var format = new FilterArguments(new StringValue("%H:%M:%S.%L"));
+            var filterFormat = new FilterArguments(new StringValue(format));
             var context = new TemplateContext { TimeZone = Eastern };
 
-            var result = await MiscFilters.Date(input, format, context);
+            var result = await MiscFilters.Date(input, filterFormat, context);
 
             Assert.Equal(expected, result.ToStringValue());
         }
--- a/Fluid/Values/FluidValue.cs
+++ b/Fluid/Values/FluidValue.cs
@@ -167,7 +167,10 @@ namespace Fluid.Values
                             return new DateTimeValue(dateTimeOffset);
 
                         case TimeSpan timeSpan:
-                            return new DateTimeValue(new DateTime(timeSpan.Ticks));
+                            // DateTimeOffset constructor throws ArgumentOutOfRangeException on locations with a positive offset
+                            // so we're resetting the offset to 0 by creating a DateTime of type UTC (Unspecified by default)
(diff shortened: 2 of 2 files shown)


sebastienros:
It doesn't feel natural that a Date created from a TimeSpan is based on UTC but one created from a Number is based on the culture's TZ. Checking on Shopify's Liquid and other implementations to understand.

danielmpetrov:
Exactly. I personally would expect all of them to respect the culture set by the user.

sebastienros:
Go with that (from what I am seeing right now). The truth is in the Number to Date conversion, seems like this in Ruby (Time.at(1)) and in LiquidJS. I assume TimeSpan should follow the same logic, only converting the TimeSpan to a number of seconds.

danielmpetrov:
This implementation matches the output of LiquidJS, the only problem is I lost the milliseconds somehow. 😂

danielmpetrov:
OK, I think this does it... let me know 👍

Fluid.Tests/MiscFiltersTests.cs
```diff
@@ -393,15 +393,15 @@ public async Task DateIsParsed()
...
-        [InlineData(0, "0")]
-        [InlineData(10, "10")]
-        [InlineData(-10, "-10")]
+        [InlineData(0, "Wed Dec 31 19:00:00.000 1969 -0500 0")]
+        [InlineData(10, "Wed Dec 31 19:00:10.000 1969 -0500 10")]
+        [InlineData(-10, "Wed Dec 31 18:59:50.000 1969 -0500 -10")]
         public async Task DateNumberIsParsedAsSeconds(long number, string expected)
         {
-            // Converting to Unix time should not vary by TimeSone
+            // Converting to Unix time should not vary by TimeZone
```
danielmpetrov, line 401: Not sure what these comments are referring to, are they still relevant?

danielmpetrov:
The problem with this is the string output for time span would be a number. I think we need a `TimeSpanValue` type, don't we?

Fluid/Values/FluidValue.cs
```diff
@@ -167,7 +167,7 @@ public static FluidValue Create(object value, TemplateOptions options)
                             return new DateTimeValue(dateTimeOffset);
 
                         case TimeSpan timeSpan:
-                            return new DateTimeValue(new DateTime(timeSpan.Ticks));
+                            return NumberValue.Create((decimal)timeSpan.TotalSeconds);
```
sebastienros, line 170: Now I remember the difference. When I added `TimeSpan` support I intended it to be converted to a `DateTimeValue`. I need to check the issue that triggered it.

But this doesn't change the question about using the current TZ or not. I will take some more time to understand the intent in shopify. TimeSpan and Number my end up having different behavior in the end.

danielmpetrov:
I added template tests to verify how a `DateTime` and `TimeSpan` object are rendered by default, without any pipes. I think these tests are missing and needed, even in the context of these changes.

Fluid/Values/FluidValueExtensions.cs
```diff
@@ -95,7 +95,8 @@ public static bool TryGetDateTimeInput(this FluidValue input, TemplateContext co
             }
             else if (input.Type == FluidValues.Number)
             {
-                var dateTime = DateTimeOffset.FromUnixTimeSeconds((long)input.ToNumberValue());
+                var milliseconds = input.ToNumberValue() * 1000;
```
sebastienros, line 98: Why did you change that?

danielmpetrov: Because the millisecond precision was lost. If the span had `1.1` seconds, it was rounded down to `1` second.

danielmpetrov: Issue is specific to this implementation, where we're treating the span as a number of seconds. But probably it doesn't make sense altogether, since now spans would be rendered as a number.

danielmpetrov:
@sebastienros Have you had time to consider how a time span should behave in Fluid? 👀

Fluid/Values/FluidValue.cs
sebastienros: Looking at Shopify's portal, when passing `now | date` or `0 | date` it's using the local TZ to create the date value. So I think we should do the same. I will update your PR to reflect this. I still want a TimeSpan to be converted to a date value though.

sebastienros:
@danielmpetrov let me know if you are ok with my changes and I'll merge it.

danielmpetrov:
@sebastienros Looks good to me, and makes sense. Glad this is resolved. 😄

PR #524, opened 2022-11-03
ReSharper ViewNotResolved error fix.

mzuvin:


hishamco:
I don't see any value in this PR

sebastienros:
I believe if you use ReSharper it has an analyzer to verify that the views you are trying to use actually exist, but in this case it doesn't find the cshtml file so it either triggering a warning or an error. This comment will tell R# to just ignore the view.

sebastienros:
There is a way to disable the feature at the editor level, but it would also disable it for other projects.

PR #530, opened 2022-11-29
Add net7.0 target

sebastienros:


hishamco:
How can react in OC while it not supports .NET 7.0 yet?!!

hishamco:
How can react in OC while it not supports .NET 7.0 yet?!!

hishamco:
How can react in OC while it not supports .NET 7.0 yet?!!

sebastienros:
???

hishamco:
Sorry, for duplicate replies, I just hide them. It happen when you write a comment from a bad internet connection ;)

What I mean is can we upgrade the package in OC while OC still not upgraded to .NET 7.0 yet?

sebastienros:
I believe the confusion here is that some projects now target net7.0 only, but these are sample projects. The main library targets netstandard (so will always work for everything) but also has specific targets for net7.0 (and older) such that it can exclude some dependencies that are in net7.0 but not in netstandard. OC uses net6.0, so it will pick the net6.0 TFM of this library, even if we "add" a net7.0 one.

PR #543, opened 2022-12-20
Implement missing strftime formats

### LINKED ISSUE #541: Time conversion problem
var source = "date:\"{{Date | date:'%F %T.%3N'}}\"";

image

%3N does not take effect , 是什么问题呢？

Of course, use format_ Date is OK
| format_date:'yyyy-MM-ddTHH:mm:ss.fffZ'

sebastienros:
Fixes #541

### DIFF
--- a/Fluid/Filters/MiscFilters.cs
+++ b/Fluid/Filters/MiscFilters.cs
@@ -1,14 +1,15 @@
-﻿using System;
+﻿using Fluid.Values;
+using System;
+using System.Buffers;
 using System.Collections.Generic;
 using System.Globalization;
-using System.Net;
-using System.Text.Json;
-using Fluid.Values;
-using TimeZoneConverter;
-using System.Threading.Tasks;
-using System.Text;
 using System.IO;
+using System.Net;
 using System.Reflection;
+using System.Text;
+using System.Text.Json;
+using System.Threading.Tasks;
+using TimeZoneConverter;
 
 namespace Fluid.Filters
 {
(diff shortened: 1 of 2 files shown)


Fluid/Filters/MiscFilters.cs
```diff
@@ -411,34 +414,60 @@ void ForStrf(DateTimeOffset value, string format, StringBuilder result)
...
                                     break;
                                 }
+                            case 'L': result.Append(Format(value.Millisecond, 3)); break;
+                            case 'm': result.Append(Format(value.Month, 2)); break;
                             case 'M':
-                                result.Append(value.Minute.ToString(context.CultureInfo).PadLeft(2, '0'));
+                                result.Append(Format(value.Minute, 2));
                                 break;
+                            case 'n': result.Append(new String('\n', width ?? 1)); break;
+                            case 'N': var v = value.Millisecond.ToString(context.CultureInfo); width ??= 9;  result.Append(v.Length >= width ? v.Substring(0, width.Value) : v.PadRight(width.Value, '0')); break;
```
nicholas-brooks, line 470: I was interested to learn how this was implemented.  On this line, Millisecond is being used, however DateTime/Offset support down to 100ns precision via ticks.

Could you perhaps use something like:

```csharp
var v = (value.Ticks % 10000000).ToString(context.CultureInfo); // get part-second from ticks in 100ns precision.
width ?? = 9;
result.Append(v.Length >= width ? v.Substring(0, width.Value) : v.PadRight(width.Value, '0'));
break;
```

sebastienros: Thanks for the suggestion. The nanoseconds are parsed to the 7th digit but that's already an improvement. Also note that net7.0 introduced `.Microseconds` and `.Nanoseconds` properties, however they don't represent what I would have expected.

Issue #578, opened 2023-06-12
Question: Introspection of parsed template


knightmeister:
Is it possible to introspect tags within a parsed template?

I know that `FluidTemplate` contains a list of `Statement`s but I don't see a way of viewing the tag name and arguments.

What I'm trying to do is look at the layout being used when rendering to do some things before giving it to Fluid to render.

i.e. 
```
{% layout "mylayout.liquid" %}
```

I would like to be able to find the tag that is specific for `layout` and then get the value `mylayout.liquid` from that tag.

Thanks :)

sebastienros:
Exactly what I am doing in a different branch, but tied to a specific feature so won't be merged soon.

When you have the list of Statements, try to find the implementation :

This tag is coming from [here](https://github.com/sebastienros/fluid/blob/8564be1aa195c31fd9b160cd0cba170d1c4f2713/Fluid.ViewEngine/FluidViewParser.cs#L73)

I believe you are now familiar of what this code does. The issue that I am seeing is that it ultimately calls into https://github.com/sebastienros/fluid/blob/8564be1aa195c31fd9b160cd0cba170d1c4f2713/Fluid/FluidParser.cs#L548 and I see that the tag name is not kept in the statement node, so you would not be able to find "layout" among all the statements of type `ParserTagStatement`. I could suggest that you customize the `FluidViewParser` instance by calling `RegisterExpressionTag("layout", ...` again but returning a Statement type that would have the name of the tag in it.

Or, you could submit a PR that add the tag information to the current class (maybe event `Statement`) and you wouldn't have to do something custom in your own code. I think this would be a good addition.

knightmeister:
Appreciate the reply. I'm happy to have a crack at doing a PR for `Statement.cs` to add the name. Will let you know how I go.

knightmeister:
PR submitted for review

MithrilMan:
> Exactly what I am doing in a different branch, but tied to a specific feature so won't be merged soon.

@sebastienros 
Is it maybe something related to the issue I opened here? https://github.com/sebastienros/fluid/issues/580

I was trying to understand the best approach to solve my problem and I thought that maybe I should use a custom parser tag but if you are working on a similar feature or something that could help solve my problem I could wait or fork your branch. But I can't find an updated branch so maybe it's private?

knightmeister:
@MithrilMan I have a PR open for this here https://github.com/sebastienros/fluid/pull/579
[...]

PR #579, opened 2023-06-14
sebastienros/fluid#578 Implement tag names

knightmeister:
I've implemented support to retrieve tag names from the statement list as discussed in #578.

`ExtensibilityTests` have been amended; let me know if I've missed anything obvious.

In terms of design, I opted to add another interface `IHasTagName` for a few reasons:

1. It means that the various statement types (`EmptyBlockStatement`, `EmptyTagStatement`, `ParserBlockStatement`, `ParserTagStatement`) now have a shared interface type
2. The above types were all `internal` and this hasn't changed visibility
3. The root class `Statement` is used throughout the codebase and wasn't an appropriate change point.

The change means that I can do what I needed to do quite simply, basically

`template.Statements.Where(x => x.GetType() == typeof(IHasTagName)).First(y => y.TagName == "what i'm looking for")`

I've also added an interface `IHasValue` for retrieving the interface value from a statement. This Value will hold an individual value passed in, or in the case of using `ArgumentsList` you'll get an `List` of `FilterArgument`.

If you need any amendments let me know.

### DIFF
--- a/Fluid.Tests/Extensibility/ExtensibilityTests.cs
+++ b/Fluid.Tests/Extensibility/ExtensibilityTests.cs
@@ -1,4 +1,5 @@
 ﻿using Fluid.Ast;
+using Fluid.Parser;
 using Parlot.Fluent;
 using System;
 using Xunit;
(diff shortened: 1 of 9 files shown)


knightmeister:
@sebastienros any feedback on this PR or plans to incorporate it?

PR #594, opened 2023-09-28
Expose own properties of scopes publicly

### LINKED ISSUE #553: Child Scope Does Not Modify Outer Scope Variables
When entering a child scope, variables defined in the parent scope can be read, but changes to their values do not propagate when the child scope is released. I would expect that changes made to a variable would persist to the outermost scope in which the variable is defined, is that correct? We are migrating to Fluid from DotLiquid and the behavior is different.

Here is an example (failing) unit test to demonstrate the issue, tested against Fluid v2.3.1:
```
        [Fact]
        public void ChildScope_Can_Modify_Outer_Scope_Variables()
        {
            var parser = new CustomParser();

            parser.RegisterExpressionBlock("repeat", async (value, statements, writer, encoder, context) =>
            {
                context.EnterChildScope();
                var repeatCount = (int)value.EvaluateAsync(context).Result.ToNumberValue();
                for (var i = 0; i < repeatCount; i++)
                {
                    await statements.RenderStatementsAsync(writer, encoder, context);
                }
                context.ReleaseScope();

                return Completion.Normal;
            });

            var template = parser.Parse("{% assign outer = 10 %}Outer={{ outer }}||{% repeat 3 %}{% assign outer = outer | plus: 1 %}Outer={{ outer }},{%- endrepeat %}||Outer={{ outer }}");
            var result = template.Render();

            Assert.Equal("Outer=10||Outer=11,Outer=12,Outer=13,||Outer=13", result);
        }
```

**Expected Output:**
Outer=10||Outer=11,Outer=12,Outer=13,||Outer=13
**Actual Output:**
Outer=10||Outer=11,Outer=12,Outer=13,||Outer=10

sebastienros:
Fixes #553

### DIFF
--- a/Fluid/Scope.cs
+++ b/Fluid/Scope.cs
@@ -7,7 +7,7 @@ namespace Fluid
 {
     public class Scope
     {
-        internal Dictionary<string, FluidValue> _properties;
+        private Dictionary<string, FluidValue> _properties;
         private readonly bool _forLoopScope;
 
         public Scope()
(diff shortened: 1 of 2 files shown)


sebastienros:
@MrUpsideDown

Fluid/Scope.cs
```diff
@@ -66,41 +66,67 @@ public FluidValue GetValue(string name)
...
             else
             {
                 Parent.SetValue(name, value);
             }
         }
 
+        /// <summary>
+        /// Sets the value with the specified name in the current scope.
+        /// </summary>
+        public void SetOwnValue(string name, FluidValue value)
```
lahma, line 116: maybe ` [MethodImpl(MethodImplOptions.AggressiveInlining)]`?

sebastienros: What's your approach with that, I am tending to let the JIT decide these days because I don't have a benchmark that proves it better. But do you think it can never be worse than not having it?

lahma: JIT can help and be wise, but probably not on full framework. I made the comment mostly because you converted from direct field access to method so call site would more likely have old behavior with aggressive inlining.

sebastienros: Good point. I assume it doesn't hurt to add it whatsoever so I am adding it back.

PR #601, opened 2023-11-01
Di/partials location formats/async writing

MoazAlkharfan:
- Use `FluidViewEngineOptions.PartialsLocationFormats` to lookup partials files
  - https://github.com/sebastienros/fluid/blob/5067057abc854e0026bfc214dbfd467776a9d60c/Fluid.ViewEngine/FluidViewRenderer.cs#L284-L307

- Use DI for `FluidViewRenderer` to allow for overriding
  - https://github.com/sebastienros/fluid/blob/5067057abc854e0026bfc214dbfd467776a9d60c/Fluid.MvcViewEngine/MvcViewFeaturesMvcBuilderExtensions.cs#L27-L31

- Configure for `FluidMvcViewOptions` to have consistent options values across the inheritance hierarchy
  -  https://github.com/sebastienros/fluid/blob/7032077df1fb3cf7b6a4acfa22222a7612271452/Fluid.MvcViewEngine/MvcViewFeaturesMvcBuilderExtensions.cs#L19-L26

```c#
services
       .AddControllersWithViews()
       .AddFluid(options => {
              /// ... Anything using FluidMvcViewOptions doesn't work with the configuration set here

              /// Options need to be repeated
       });

// Need this to set mvc options
services
       .Configure<FluidMvcViewOptions>(x =>
       {
              /// Before:
              /// Options need to be repeated here

              /// After:
              // Only this is needed to be configured
              x.RenderingViewAsync = (path, viewContext, context) =>
              {
                     return ValueTask.CompletedTask;
              };
       });
```

- Add `WriteToAsync` because otherwise kestrel would throw and requires to override `AllowSynchronousIO`
```c#
builder.WebHost
        .UseKestrel(p =>
        {
            p.AllowSynchronousIO = true;
        });
```


sebastienros:
@MoazAlkharfan I have integration the async part of your PR in https://github.com/sebastienros/fluid/pull/667, thanks for the contribution. I am keeping this one open since it also contains the configuration improvements but haven't looked at it closely enough to understand the problem so far (but I trust you about this).

PR #611, opened 2023-12-08
Seal types not suitable for extending

lahma:
CLR can make better decisions when it knows there's no worry to do virtual calls to ensure that possible overridden method would be correctly called. Types really don't have extension points via inheritance.

## Fluid.Benchmarks.FluidBenchmarks

| **Diff**|Method|Mean|Error|Allocated|
|------- |-------|-------:|-------|-------:|
| Old |Parse|3.193 μs|0.0312 μs|2.68 KB|
| **New** |	| **3.152 μs (-1%)** | **0.0307 μs** | **2.68 KB (0%)** |
| Old |ParseBig|16.892 μs|0.2980 μs|11.61 KB|
| **New** |	| **16.840 μs (0%)** | **0.2747 μs** | **11.61 KB (0%)** |
| Old |Render|133.681 μs|2.5941 μs|95.86 KB|
| **New** |	| **130.399 μs (-2%)** | **1.4315 μs** | **95.86 KB (0%)** |
| Old |ParseAndRender|135.980 μs|1.5126 μs|99.01 KB|
| **New** |	| **135.876 μs (0%)** | **2.0597 μs** | **99.01 KB (0%)** |

### DIFF
--- a/Fluid/Accessors/AsyncDelegateAccessor.cs
+++ b/Fluid/Accessors/AsyncDelegateAccessor.cs
@@ -3,7 +3,7 @@ using System.Threading.Tasks;
 
 namespace Fluid.Accessors
 {
-    public class AsyncDelegateAccessor : AsyncDelegateAccessor<object, object>
+    public sealed class AsyncDelegateAccessor : AsyncDelegateAccessor<object, object>
     {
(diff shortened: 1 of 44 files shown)


sebastienros:
I fear that it will break someone. But let's try and see people complain, the current version being quite stable they shouldn't need to upgrade and won't be blocked.

lahma:
This was my thinking too, let's break it and unseal when needed.

PR #612, opened 2023-12-08
Handle IConvertible case to return specific type FluidValues

### LINKED ISSUE #609: Using json filter on a JObject coverts all values to string
When passing a JObject as the context and using the `json` filter in the template, all the values, regardless of the original datatype, are getting converted to string in the output. 

For example, consider the following program:
```csharp
using Fluid;
using Newtonsoft.Json.Linq;

var template = parser.Parse("{{ root | json: true }}");

var contextObj = new JObject
{
    ["root"] = new JObject
    {
        ["a"] = true,
        ["b"] = 1,
        ["c"] = DateTime.Now,
        ["d"] = "string",
        ["e"] = null,
        ["f"] = new JObject
        {
            ["f_a"] = 1.2
        },
        ["g"] = new JArray
        {
            "val1", "val2"
        }
    }
};
var context = new TemplateContext(contextObj);
var output = template.Render(context);
Console.WriteLine(output);
```
The output of the program is:
```
{
    "a": "True",
    "b": "1",
    "c": "12/08/2023 20:00:38",
    "d": "string",
    "e": "",
    "f": {
        "f_a": "1.2"
    },
    "g": [
        "val1",
        "val2"
    ]
}
```
All the basic datatypes are getting converted to string, which is not correct.

The expected output of the program is:
```
{
    "a": true,
    "b": 1,
    "c": "12/08/2023 20:00:38",    // Not sure what is the ideal way to print date time objects
    "d": "string",
    "e": null,
    "f": {
        "f_a": 1.2
    },
    "g": [
        "val1",
        "val2"
    ]
}
```

atthevergeof:
fixes #609 

The `JValue` class does implement IConvertible, but in `Fluid.Create`, when `System.Type.GetTypeCode()` is called with JValue type in the argument, it returns `TypeCode.Object`. And under this case, it was going to the `IFormattable` case and being converted to string regardless of the underlying data type. But when `JValue` is cast to `IConvertible` and then the `GetTypeCode()` method is called on the object, it does return the correct type code corresponding to the underlying value. 

So to resolve the issue #609, I have moved the `IConvertible` case above `IFormattable`, and also changed its implementation to extract the underlying value and call the Create method again with that value. To prevent any infinite recursion, I have explicitly handled the `TypeCode.Object` case by returning the string value for it.


lahma:
You probably want to add some test cases to demonstrate what you have fixed and also to ensure nobody breaks the behavior you are after later on.

PR #618, opened 2024-01-05
Added Sum ArrayFilter with 2 different options

cpriestland:
Added "Sum" ArrayFilter so that others can use it also.

2 options:

- Can include an argument to indicate which member to Sum on an array of objects
- No argument which will attempt to Sum the array items value

Tests included with Pull Request which check both of these scenario's

We use Fluid quite a lot at TemplateTo and hope you don't mind us putting our link in your UsedBy section, Fluid can be used by our users as well when building out their templates.

### DIFF
--- a/Fluid.Tests/ArrayFiltersTests.cs
+++ b/Fluid.Tests/ArrayFiltersTests.cs
@@ -3,6 +3,8 @@ using Fluid.Values;
 using Fluid.Filters;
 using Xunit;
 using System.Threading.Tasks;
+using Fluid.Ast;
+using Fluid.Tests.Extensibility;
 
 namespace Fluid.Tests
 {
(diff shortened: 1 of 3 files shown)


hishamco (approved):

cpriestland:
@hishamco Thanks for approving this, is there an idea of when this could be merged as we require the functionality? Thanks

sebastienros:
Would you mind adding the same tests as in Shopify Liquid to ensure everything behaves as the specs?
https://github.com/Shopify/liquid/blob/f5ed5404b5ba04c7010d4f37ca2928bb6fa925a2/test/integration/standard_filter_test.rb#L952-L1052

cpriestland:
Thanks for your reply @sebastienros I have added the tests as suggested

hishamco:
Thanks for your contribution @cpriestland

Issue #625, opened 2024-02-01
Minimal api sample doesn't work when run on linux


AlexGirardDev:
if you try and run the minimal api sample on linux it will return a 404.

the issue is caused by the index name being set to "Index" but the liquid file is named "index.liquid"

PR #626, opened 2024-02-01
changing view name to match file casing in minimal api sample

AlexGirardDev:
fix for #625


AlexGirardDev:
the other solution to this would have been to make [this](https://github.com/AlexGirardDev/fluid/blob/529ad95a6078268d28bdd91c789e50945049a153/MinimalApis.LiquidViews/ActionViewResult.cs#L57) ignore file name casing when looking for files but just updating the view name seemed simpler

sebastienros:
I checked the other files, and I see that `_layout.liquid` is referenced using `_Layout`, so it shouldn't work either. Can you confirm?

sebastienros:
I kind of like the PascalCasing of files, if you do too then let's use that solution instead. If you don't then I'll agree with your changes.

AlexGirardDev:
good point, i just checked and layout is'nt being loaded at all because the viewstart is not being loaded.
its file name is coming from [here](https://github.com/sebastienros/fluid/blob/a24b0330f5e1140eab3909b85192e4bae00e9bff/Fluid.ViewEngine/Constants.cs#L11) which seems to be used by the renderer so i wonder if its just easier to proper case all the luiqud files

sebastienros:
Looks like we have to then, good catch.

sebastienros:
I wonder is ASP.NET is also forcing Viewstart files to follow this case on Linux.

AlexGirardDev:
i was curious and it just seemms like there file names are just completly case insensitive ASP.NET

i just started renaming these and everything worked as expected
image

they all did start as proper case though

sebastienros:
I assume you tried this on Linux? (I would think so since this is how you found about this issue)

AlexGirardDev:
Ya, everything has been on linux with Rider.

I think views and more specfically _ViewStart being case sensitive could catch some people off gaurd and may be worth calling out in the docs. Especially if they are developing on windows and then deploying to linux through docker (which is pretty common nowadays i believe).

The test i did with ASP.NET was with compiled templates so i'm assuming this is how they get around the case sensitivity thing, they can just load all there templates into memory and access they view names while ignoring case through managed c# code.

But if your loading the files from disk at runtime, I think its tricker beacuse the native OS calls on a case sesenitve system won't let you access a file while ignoring case.

I was googling around a bit and found this issue on the dotnet runtime that is tangentually related 
https://github.com/dotnet/runtime/issues/91250

Issue #551, opened 2023-01-16
Including external macros in separate .liquid partial file seems not to work


thomas-baumgarth:
If I use macros directly in the same liquid file where I invoke the macros, all macros get executed. If I put the macros in separate .liquid file and import them via the include tag, they are not executed for some reason. 
I checked if the separate liquid file is found and read by the fluid template engine and that works fine.

hishamco:
Could you please provide a sample for what you did? So, we can easily reproduce the issue - if there's -

thomas-baumgarth:
Here is the sample. Please check out the ReadMe.md
[FluidIncludeBug.zip](https://github.com/sebastienros/fluid/files/10443805/FluidIncludeBug.zip)

hishamco:
Is it in a public repo? Then write the steps to reproduce the issue

thomas-baumgarth:
?? 
I uploaded a complete sample (code and templates). 
- Please unzip the uploaded solution and run the code. 
- In the ReadMe.md I described what to do to see the error.

hishamco:
Ok, you know public repos might be TRUSTED in terms of security ;)

gebolze:
I just had the same issue while I evaluated fluid. I've created a fork and committed a failing unit-test to demonstrate the issue. https://github.com/gebolze/fluid/commit/dacd8d639e72018514b2c0c0ae1cadbf234161b9

SJFriedl:
This is certainly a surprise; the workaround Codex found is:

```
{% from 'my-macro-file.liquid' import my_macro_name %}
```

There doesn't appear to be any kind of wildcard for the import, which is unfortunate.

EDIT: `{% from 'my-macro-file.liquid' %}` does in fact work - thanks for this!

sebastienros:
@SJFriedl the pr https://github.com/sebastienros/fluid/pull/989 adds documentation about how to do that. I also closed the issue since this was fixed a while ago.

PR #632, opened 2024-03-13
Fix included macros aren't working

gebolze:
Fix for #551 

Macros included using the `include` statement are always undefined because they are defined within the scope of the `include` statement. To prevent this, this pull-request changes the implementation of the `macro` statement to always register the macro to the root scope.


gebolze:
I've updated the PR. I've changed the implementation and introduced a new `from` tag that allows importing macros from external templates.

`{% from 'file' %}` imports all defined macros.
`{% from 'file' import foo, bar %}` imports only foo, bar.

I've implemented the TagStatement such that it doesn't render anything to the output and only import macros to the local scope. This is a bit opinionated, but I think the `from` tag should not be used to for scenarios where `include` or `render` could be used.

Fluid/Ast/FromStatement.cs
```diff
@@ -0,0 +1,114 @@
...
+                {
+                    _semaphore.Release();
+                }
+            }
+
+            try
+            {
+                var parentScope = context.LocalScope;
+
+                context.EnterChildScope();
```
sebastienros, line 77:
```suggestion
                // Create a dedicated scope so we can list all macros defined in this template
                context.EnterChildScope();
```

Fluid/Ast/FromStatement.cs
```diff
@@ -0,0 +1,114 @@
...
+                await _cachedTemplate.Template.RenderAsync(TextWriter.Null, encoder, context);
+
+                if (Functions != null)
+                {
+                    foreach (var functionName in Functions)
+                    {
+                        var value = context.LocalScope.GetValue(functionName);
+                        if (value.Type != FluidValues.Nil)
```
sebastienros, line 85:
```suggestion
                        if (value is FunctionValue)
```

Like the block below?

README.md
```diff
@@ -934,7 +934,18 @@ Now `field` is available as a local property of the template and can be invoked
 {{ field('pass', type='password') }}
 ```
 
-> Macros need to be defined before they are used as they are discovered as the template is executed. They can also be defined in external templates and imported using the `{% include %}` tag.
```
sebastienros, line 937: The PR is not changing the behavior wrt includes, so this will still work, right? Not saying this should be removed. I also remember that `include` is not recommended in Liquid (because like in this case it leaks to the caller).

sebastienros (commented):
I was not expecting anybody to be able to do this without asking more questions or saying "it's too complicated", good job!

Give me some time to look again into Jinja and if their way for naming macros has advantages. I don't think it would be more complex than what you implemented, just checking the user cases and how simple the usage is.

sebastienros:
> This is a bit opinionated, but I think the from tag should not be used to for scenarios where include or render could be used.

Are you talking about not rendering the imported template? Is so then yes definitely should not render, just import functions from it.

sebastienros:
Done with checking Jinja again. Can you confirm that what you did is compliant with Jinja's syntax, but you just don't handle the `as` keywork either in the `from` or the `import` expressions? But this could be added later.

Examples:

`{% from 'forms.html' import input as input_field, textarea %}` -> imports `input` but names it `input_field` locally, then imports `textarea`

`{% import 'forms.html' as forms %}` -> `imports` all function but as `forms.input` and `forms.textarea`.

README.md
gebolze: >The PR is not changing the behavior wrt includes, so this will still work, right?
Yes behavior of includes isn't changed by this PR. But I would still say that include can't be used to add external macros to the local scope. The following example won't work:

functions.liquid
```
{%- macro hello_world() -%}
Hello world!
[%- endmacro -%}
```

template.liquid
```
{%- include 'functions' %}
{{ hello_world() }}
```

This will **not** render `Hello world!`

Actually the opposite will work.
You can access a marco declared in a template from the included file.

Hope this answers your question.

PR #645, opened 2024-04-23
fix: Directory.Packages.props does not respect PrivateAssets property

kaqq:
https://github.com/sebastienros/fluid/issues/646


lahma:
I would suggest moving them to be `GlobalPackageReference` so they will implicitly be private assets, also no need to reference them in projects either after that.

https://github.com/quartznet/quartznet/blob/72fce9c599e020f4db0b0c77d277f9971d947a6f/Directory.Packages.props#L69-L74

lahma (approved):
👍🏻

sebastienros:
Thanks @kaqq , I just published a new release with your fix.

hishamco:
Ths build is failed!!

PR #650, opened 2024-05-17
Add visitors support

sebastienros:


sebastienros:
@lahma I am merging it because someone needs the feature, but feel free to comment if you have anything to say and I'll react

hishamco:
Screenshot 2024-05-18 011441

Why?!!

sebastienros:
I assume the config is not the same for PRs and main. I am not worried, it didn't prevent the nuget publish step. We'll fix it eventually.

PR #668, opened 2024-05-29
Improve TextSpan and For statements

sebastienros:
Extracted from the compilation branch

### DIFF
--- a/Fluid/Ast/ForStatement.cs
+++ b/Fluid/Ast/ForStatement.cs
@@ -5,7 +5,6 @@ namespace Fluid.Ast
 {
     public sealed class ForStatement : TagStatement
     {
-        private readonly bool _isContinueOffset;
         private readonly string _continueOffsetLiteral;
 
         public ForStatement(
@@ -25,7 +24,7 @@ namespace Fluid.Ast
             Reversed = reversed;
             Else = elseStatement;
 
-            _isContinueOffset = Offset is MemberExpression l && l.Segments.Count == 1 && ((IdentifierSegment)l.Segments[0]).Identifier == "continue";
+            OffSetIsContinue = Offset is MemberExpression l && l.Segments.Count == 1 && ((IdentifierSegment)l.Segments[0]).Identifier == "continue";
             _continueOffsetLiteral = source is MemberExpression m ? "for_continue_" + ((IdentifierSegment)m.Segments[0]).Identifier : null;
         }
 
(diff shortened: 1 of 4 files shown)


Fluid/Ast/ForStatement.cs
```diff
@@ -36,6 +35,7 @@ public sealed class ForStatement : TagStatement
         public Expression Offset { get; }
         public bool Reversed { get; }
         public ElseStatement Else { get; }
+        public bool OffSetIsContinue { get; }
```
hishamco, line 38:
```suggestion
        public bool OffsetIsContinue { get; }
```

Fluid/Ast/ForStatement.cs
```diff
@@ -25,7 +24,7 @@ public sealed class ForStatement : TagStatement
             Reversed = reversed;
             Else = elseStatement;
 
-            _isContinueOffset = Offset is MemberExpression l && l.Segments.Count == 1 && ((IdentifierSegment)l.Segments[0]).Identifier == "continue";
+            OffSetIsContinue = Offset is MemberExpression l && l.Segments.Count == 1 && ((IdentifierSegment)l.Segments[0]).Identifier == "continue";
```
hishamco, line 27:
```suggestion
            OffsetIsContinue = Offset is MemberExpression l && l.Segments.Count == 1 && ((IdentifierSegment)l.Segments[0]).Identifier == "continue";
```

Fluid/Ast/ForStatement.cs
```diff
@@ -55,7 +55,7 @@ public override async ValueTask<Completion> WriteToAsync(TextWriter writer, Text
             var startIndex = 0;
             if (Offset is not null)
             {
-                if (_isContinueOffset)
+                if (OffSetIsContinue)
```
hishamco, line 58:
```suggestion
                if (OffsetIsContinue)
```

Fluid/Ast/ForStatement.cs
hishamco, line 27: Can we used `static` field for "Continue" identifier

Fluid/Ast/TextSpanStatement.cs
```diff
@@ -1,16 +1,15 @@
...
 {
     public sealed class TextSpanStatement : Statement
     {
-        private bool _isStripped;
-        private bool _isEmpty;
+        private bool _isBufferPrepared;
         private readonly object _synLock = new();
         private TextSpan _text;
-        private string _buffer;
+        internal string _preparedBuffer;
```
hishamco, line 12: Why `internal`?

hishamco (changes requested):

Fluid/Ast/TextSpanStatement.cs
sebastienros: It's used internally in the branch it's coming from, will keep it this way for now.

Fluid/Ast/ForStatement.cs
sebastienros: If it was used somewhere else it could be a const, but it just local so it doesn't matter.

PR #675, opened 2024-07-07
Update README.md: Add Mailgen to the list of projects using Fluid

hsndmr:
Mailgen:
A Node.js package that generates clean, responsive HTML e-mails for sending transactional mail.


hishamco:
Thanks Hasan

PR #681, opened 2024-07-12
Added `StringComparer` property to `TemplateContext`

gumbarros:
This PR allows case insensitivity at templates, making more easy to non-programmers to use variables and functions.

### DIFF
--- a/Fluid.Tests/TemplateContextTests.cs
+++ b/Fluid.Tests/TemplateContextTests.cs
@@ -164,8 +164,7 @@ namespace Fluid.Tests
         [Fact]
         public void ShouldUseStringComparer()
         {
-            var context = new TemplateContext();
-            context.StringComparer = StringComparer.InvariantCultureIgnoreCase;
+            var context = new TemplateContext(StringComparer.InvariantCultureIgnoreCase);
             context.SetValue("PageState", "insert");
 
             Assert.Equal("insert", context.GetValue("pageState").ToStringValue());
--- a/Fluid.Tests/TemplateContextTests.cs
+++ b/Fluid.Tests/TemplateContextTests.cs
@@ -165,7 +165,7 @@ namespace Fluid.Tests
         public void ShouldUseStringComparer()
         {
             var context = new TemplateContext();
-            context.StringComparer = StringComparer.InvariantCultureIgnoreCase;
+            context.Options.StringComparer = StringComparer.InvariantCultureIgnoreCase;
             context.SetValue("PageState", "insert");
 
             Assert.Equal("insert", context.GetValue("pageState").ToStringValue());
--- a/Fluid.Tests/TemplateContextTests.cs
+++ b/Fluid.Tests/TemplateContextTests.cs
@@ -162,9 +162,19 @@ namespace Fluid.Tests
         }
         
         [Fact]
-        public void ShouldUseStringComparer()
+        public void ShouldUseCustomStringComparer()
         {
-            var context = new TemplateContext(StringComparer.InvariantCultureIgnoreCase);
+            var context = new TemplateContext(StringComparer.OrdinalIgnoreCase);
+            context.SetValue("PageState", "insert");
+
+            Assert.Equal("insert", context.GetValue("pageState").ToStringValue());
+        }
+
+        [Fact]
+        public void ShouldUseTemplateOptionsStringComparer()
+        {
+            var options = new TemplateOptions { ModelNamesComparer = StringComparer.OrdinalIgnoreCase };
+            var context = new TemplateContext(options);
             context.SetValue("PageState", "insert");
 
             Assert.Equal("insert", context.GetValue("pageState").ToStringValue());
--- a/Fluid.Tests/TemplateContextTests.cs
+++ b/Fluid.Tests/TemplateContextTests.cs
@@ -167,6 +167,7 @@ namespace Fluid.Tests
             var context = new TemplateContext();
             context.StringComparer = StringComparer.InvariantCultureIgnoreCase;
             context.SetValue("PageState", "insert");
+
             Assert.Equal("insert", context.GetValue("pageState").ToStringValue());
         }
 
--- a/Fluid.Tests/TemplateContextTests.cs
+++ b/Fluid.Tests/TemplateContextTests.cs
@@ -165,7 +165,7 @@ namespace Fluid.Tests
         public void ShouldUseStringComparer()
         {
             var context = new TemplateContext();
-            context.Options.StringComparer = StringComparer.InvariantCultureIgnoreCase;
+            context.StringComparer = StringComparer.InvariantCultureIgnoreCase;
             context.SetValue("PageState", "insert");
 
             Assert.Equal("insert", context.GetValue("pageState").ToStringValue());
--- a/Fluid.Tests/TemplateContextTests.cs
+++ b/Fluid.Tests/TemplateContextTests.cs
@@ -160,6 +160,15 @@ namespace Fluid.Tests
 
             Assert.Equal("101", output);
         }
+        
+        [Fact]
+        public void ShouldUseStringComparer()
+        {
+            var context = new TemplateContext();
+            context.StringComparer = StringComparer.InvariantCultureIgnoreCase;
+            context.SetValue("PageState", "insert");
+            Assert.Equal("insert", context.GetValue("pageState").ToStringValue());
+        }
 
         private class TestClass
         {
--- a/Fluid/Scope.cs
+++ b/Fluid/Scope.cs
@@ -15,7 +15,7 @@ namespace Fluid
         public Scope(Scope parent)
         {
             Parent = parent;
-            StringComparer = parent.StringComparer;
+            StringComparer = parent?.StringComparer ?? StringComparer.CurrentCulture;
         }
 
         public Scope(Scope parent, bool forLoopScope)
--- a/Fluid/Scope.cs
+++ b/Fluid/Scope.cs
@@ -15,7 +15,6 @@ namespace Fluid
         public Scope(Scope parent)
         {
             Parent = parent;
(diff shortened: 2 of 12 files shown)


Fluid.Tests/TemplateContextTests.cs
```diff
@@ -160,6 +160,15 @@ public async Task ShouldNotReleaseScopeAsynchronously()
...
             Assert.Equal("101", output);
         }
+        
+        [Fact]
+        public void ShouldUseStringComparer()
+        {
+            var context = new TemplateContext();
+            context.StringComparer = StringComparer.InvariantCultureIgnoreCase;
+            context.SetValue("PageState", "insert");
+            Assert.Equal("insert", context.GetValue("pageState").ToStringValue());
```
hishamco, line 170:
```suggestion

            Assert.Equal("insert", context.GetValue("pageState").ToStringValue());
```

Fluid/Scope.cs
```diff
@@ -7,28 +7,32 @@ public sealed class Scope
...
+            StringComparer = parent.StringComparer;
         }
 
         public Scope(Scope parent, bool forLoopScope)
         {
             if (forLoopScope && parent == null) ExceptionHelper.ThrowArgumentNullException(nameof(parent));
 
             Parent = parent;
-
+
```
hishamco, line 29:
```suggestion

```

Fluid/Scope.cs
```diff
@@ -7,28 +7,32 @@ public sealed class Scope
...
-
+            
             // A ForLoop scope reads and writes its values in the parent scope.
             // Internal accessors to the inner properties grant access to the local properties.
             _forLoopScope = forLoopScope;
-
-            _properties = new Dictionary<string, FluidValue>();
+            StringComparer = parent.StringComparer;
+            _properties =
+                new Dictionary<string, FluidValue>(StringComparer);
```
hishamco, line 35:
```suggestion
            _properties = new Dictionary<string, FluidValue>(StringComparer);
```

Fluid/Scope.cs
```diff
@@ -7,28 +7,32 @@ public sealed class Scope
     {
         private Dictionary<string, FluidValue> _properties;
         private readonly bool _forLoopScope;
+        public StringComparer StringComparer { get; set; }
```
hishamco, line 10: Move it after the constructors

hishamco (changes requested):

Fluid/TemplateContext.cs
```diff
@@ -150,6 +150,15 @@ public void IncrementSteps()
         /// </summary>
         public Func<string, string, ValueTask<string>> Captured { get; set; }
 
+        /// <summary>
+        /// Gets or sets the <see cref="StringComparer"/> used for matching keys of scope values.
+        /// </summary>
+        public StringComparer StringComparer
```
hishamco, line 156: I think you can use the one that is declared in the `TemplateOptions`

https://github.com/sebastienros/fluid/blob/3f1750f690b5b2e02cd13e6554e0df832dbd8385/Fluid/TemplateContext.cs#L85

Fluid/TemplateContext.cs
```diff
@@ -149,7 +149,7 @@ public void IncrementSteps()
         /// Gets or sets the delegate to execute when a Capture tag has been evaluated.
         /// </summary>
         public Func<string, string, ValueTask<string>> Captured { get; set; }
-
+
```
hishamco, line 152:
```suggestion
```

Fluid/TemplateOptions.cs
```diff
@@ -73,6 +73,14 @@ public class TemplateOptions
         /// </summary>
         public bool Greedy { get; set; } = true;
 
+        /// <summary>
+        /// Gets or sets the <see cref="StringComparer"/> used for matching keys of scope values.
+        /// </summary>
+        public StringComparer StringComparer
```
hishamco, line 79: Where this property is used?

hishamco (changes requested):

Fluid/TemplateOptions.cs
gumbarros: At my unit test to change the `StringComparer` all way down to the scope.

gumbarros: Wait, my test is failing because the property was removed.

hishamco: Can I fix this?

gumbarros: > Can I fix this?
Yes

hishamco: I will check ..

hishamco:
@gumbarros check my last commit

gumbarros:
> @gumbarros check my last commit

LGTM

Fluid/Scope.cs
```diff
@@ -11,25 +11,31 @@ public sealed class Scope
         public Scope()
         {
             Parent = null;
+            StringComparer = StringComparer.CurrentCulture;
         }
```
hishamco, line 15:
```suggestion
        public Scope() : this(null)
        {
        }
```

hishamco (changes requested):

hishamco (approved):

hishamco:
@sebastienros anything else to add to this?

hishamco:
Thanks for a quick fix :)

gumbarros:
Any plans to merge this 👀?

gumbarros:
I'm going to be out for 20 days next week and won't be able to make any changes for this PR. I would also like to use this feature this week if possible.

hishamco:
@sebastienros any feedback on this or shall I merge?

sebastienros:
This is too intrusive (in the ctor). Can this `StringComparer` be yet another property of the options/context classes, be inherited like the other properties?

And the new property should be named `ModelNamesComparer`? The default value should be Ordinal obviously.

sebastienros:
@gumbarros 

if you can't do the changes @hishamco or myself should be able to do it.

hishamco:
To be honest it was as it's seen in https://github.com/sebastienros/fluid/pull/681/commits/b99e0f54853fbebf42101758f24388917e2ac32f

If you like how it was I will revert the latest changes considering the name that you suggested

sebastienros:
> To be honest it was as it's seen in https://github.com/sebastienros/fluid/commit/b99e0f54853fbebf42101758f24388917e2ac32f

Did I actually ask the opposite, or was it done on your own thoughts?

hishamco:
It added as property at the beginning but I saw the property appears twice in different places that's why I suggested to pass the values through constructor

sebastienros:
In you case that's what you mean, it's pretty common to see the properties duplicated and cloned from `TemplateOptions` and `TemplateContext`. But this is done this way for anything that can alter the behavior or the rendering, to have defaults or options to share across template content instances without having to set it each time.

hishamco:
I see if the linked commit is fine for you I will revert the latest changes then merge unless you need to add something else

gumbarros:
Hi guys, any news on merging this 👀?

sebastienros:
Updated it. Should be simpler, and it helped me realize that this property can't be changed dynamically as it is used by an inner collection.

In Fluid 3.0 we might want to use the same comparer for both internal properties and model properties. (note for myself)

nguyenngocanhpro:
When will this PR be released? I've been waiting for so long. 
And I think the `GetValueAsync` function from `FluidValue` should be updated in the same way as this case

sebastienros:
@nguyenngocanhpro do you want to submit a PR for GetValueAsync? I can ship it right after that.

nairdo:
With this PR (tested using v2.24.0), should we still be able to use **case-sensitive** variables if the `TemplateOptions ModelNamesComparer` is set to `StringComparer.Ordinal`?  

I was under the impression that Shopify Liquid still supports this behavior ([example](https://liquidjs.com/playground.html#IHslIGFzc2lnbiBjYXNlID0gImxvd2VyIiAlfSANCiB7JSBhc3NpZ24gQ0FTRSA9ICJ1cHBlciIgJX0gDQogY2FzZSA9IHt7IGNhc2UgfX0gPCEtLSBleHBlY3QgbG93ZXIgLS0+DQogQ0FTRSA9IHt7IENBU0UgfX0gPCEtLSBleHBlY3QgdXBwZXIgLS0+,ewogICJwZW9wbGUiOiBbCiAgICAiYWxpY2UiLAogICAgImJvYiIsCiAgICAiY2Fyb2wiCiAgXQp9Cg==)):

```Liquid
 {% assign case = "lower" %} 
 {% assign CASE = "upper" %} 
 case = {{ case }} <!-- expect lower -->
 CASE = {{ CASE }} <!-- expect upper -->
 ```
 
 Here's a unit test:
 ```CSharp
[Fact]
public void ShouldUseTemplateOptionsStringComparerWithCaseSensitive()
{
    var options = new TemplateOptions { ModelNamesComparer = StringComparer.Ordinal };
    var context = new TemplateContext(options);
    context.SetValue("case", "lower");
    context.SetValue("CASE", "upper");
    context.SetValue("Case", "mixed");

    Assert.Equal("lowerupper", context.GetValue("case").ToStringValue() + context.GetValue("CASE").ToStringValue());
}
 ```
 
 Results:

`     Actual:   "mixedmixed"`

sebastienros:
The link you provided is from liquidjs, not shopify, here is the result from Shopify (https://shopify.dev/docs/api/liquid)

image

nairdo:
@sebastienros Right, sorry about using liquidjs, I just mainly wanted to confirm that Fluid _should_ still allow case sensitivity.  I'll open this as a new issue.

sebastienros:
I just wanted to be sure it was also the behavior in shopify. I also create issues on liquidjs because it doesn't follow the standard so I wanted to point it out.

sebastienros:
Here is the PR fixing the bug: https://github.com/sebastienros/fluid/pull/802

PR #691, opened 2024-08-22
Trim = from Base64UrlSafeEncode

### LINKED ISSUE #690: Base64UrlSafeEncode should remove `==`
I believe this should do it as `=` is not "url safe" and is just used in Base64 for padding -> `TrimEnd('=')`

Also there is a new class to do it in .NET 9, we may want to use it when available:

```
#if NET9_0_OR_GREATER
        return Base64Url.EncodeToString(input);
#else
```

hishamco:
Fixes #690

### DIFF
--- a/Fluid/Filters/MiscFilters.cs
+++ b/Fluid/Filters/MiscFilters.cs
@@ -196,7 +196,7 @@ namespace Fluid.Filters
                 encodedBase64StringBuilder.Replace('+', '-');
                 encodedBase64StringBuilder.Replace('/', '_');
 
-                return new StringValue(encodedBase64StringBuilder.ToString());
+                return new StringValue(encodedBase64StringBuilder.ToString().TrimEnd('='));
             }
         }
 


Fluid/Filters/MiscFilters.cs
```diff
@@ -196,7 +196,7 @@ public static ValueTask<FluidValue> Base64UrlSafeEncode(FluidValue input, Filter
                 encodedBase64StringBuilder.Replace('+', '-');
                 encodedBase64StringBuilder.Replace('/', '_');
 
-                return new StringValue(encodedBase64StringBuilder.ToString());
+                return new StringValue(encodedBase64StringBuilder.ToString().TrimEnd('='));
```
sebastienros, line 199: Does the string builder have something like it? That would prevent the extra string allocation.

sebastienros (commented):
Can you add to the unit tests?

Hell -> SGVsbA== -> SGVsbA
Hello -> SGVsbG8= -> SGVsbG8

sebastienros:
Feel free to merge when you add the test, thanks.

Fluid/Filters/MiscFilters.cs
lahma: If we know that base64 encoded version ends with `==` (it should) it should suffice to `encodedBase64StringBuilder.Length -= 2;`

hishamco: That's a good option too

sebastienros: It's not always the case, could be zero, one or two chars. See the test cases I shared for instance.

sebastienros: But I see what you mean now, sorry.

hishamco: The issue the might be raise during decoding how we know there was zero, one or two "=" on the encoded URL?

hishamco: FYI there's `Base64.IsValid()` introduced in .NET 8.0 but it will not solve the problem in our case

sebastienros:
Added the same tests to ensure the decoding was fine too and it happens it was not as the `=` chars are required by the API.

Took the time to implement @lahma 's suggestion to skip the string allocation.

Issue #685, opened 2024-07-24
Cannot render nested anonymous objects returned from custom filter


arthurmmedeiros:
Hello there,

First of all I would like to thank for all the work that has been done in this library. I've been using it for a while, and it has been really useful in some of my projects.

**Problem:**
I am currently facing a problem with trying to write custom filter. I wanted to implement a custom filter that performs a `group by` and returns a list of grouped objects. I got it sort of working, the problem is that I cannot get it to access the nested objects inside the grouped lists. I can for instance access they key value of the objects, but the list of items seems to be ignored

To illustrate the problem, I have a template that looks like this
`"<h1>Hello</h1><div>  {% assign events_by_type = events | group_by: \"eventTypeId\" %}  {% for type_group in events_by_type %} <div> {{type_group.key}}  {% for item in type_group.items %} {{ item.name }} {{item}} {% endfor %} </div> {% endfor %} </div>"`

and I am using a json object that looks like this:
`"{\"events\":[{\"eventTypeId\":1234,\"type\":{\"id\":1234,\"name\":\"type 1234\"}}, {\"eventTypeId\": 1111, \"type\":{\"id\":1111,\"name\":\"type 1111\"} }]}"`

This is the result I get 
`<h1>Hello</h1><div>     <div> 1234      </div>  <div> 1111      </div>  </div>`

This is what I was expecting
`<h1>Hello</h1><div>     <div> 1234   type 1234   </div>  <div> 1111   type 1111   </div>  </div>`
anonymous objects
I have a json object containing all the data I want to render. The catch with this json object is that I cannot register my classes as instructed [here](https://github.com/sebastienros/fluid#allow-listing-object-members). I want it to be generic as possible so I can work with any json object.

For this to happen I thought all I needed to do was to use `UnsafeMemberAccessStrategy`, but it does not seem to be the case.

Here is a simple repro [https://dotnetfiddle.net/r0hkwE](https://dotnetfiddle.net/r0hkwE)

I want to be able to do group by using metadata and by simple properties as well, hence why I do `ToStringValue().Split(".")`. 

I don't believe this is an issue with the custom filter itself since I can access the key values inside the list of objects. I cannot access the list of items inside the objects though. 

I'm not sure if this could be a bug or if it's me.

arthurmmedeiros:
I realised my code was all wrong. Adde this [PR](https://github.com/sebastienros/fluid/pull/692) with the feature.

PR #692, opened 2024-08-26
Add custom filter group_by

arthurmmedeiros:
Hello there,

I previously added this [question](https://github.com/sebastienros/fluid/issues/685), but I realised the code I wrote was just wrong. I fixed it and decided to create a pull request with this new custom filter. 

I think this is a useful feature, so I felt I could share and give a small contribution to this library.

I added some tests and I'm happy to add more if necessary.

Thank you.

### DIFF
--- a/Fluid.Tests/ArrayFiltersTests.cs
+++ b/Fluid.Tests/ArrayFiltersTests.cs
@@ -4,6 +4,7 @@ using Fluid.Filters;
 using Xunit;
 using System.Threading.Tasks;
 using Fluid.Tests.Extensibility;
+using Newtonsoft.Json.Linq;
 
 namespace Fluid.Tests
 {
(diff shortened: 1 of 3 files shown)


hishamco:
What's the problem that you going to fix?

arthurmmedeiros:
@hishamco sorry, I should have given more details about this pull request in my first comment.

The problem I am trying to solve is to group a list of objects by a certain key. In the tests I used an example where I had a list of users, and I wanted to group these users by type (admin and client).

I hope this makes more sense.

hishamco:
@sebastienros do you have any objection to introduce this filter, before I review

yegorandrosov:
grouping would also be useful for our use case

especially some complex cases like
{% assign osVersions = model.Variants | group_by: 'CustomFields["97f9af79-a4ea-4766-8eac-b61da5d89de7"].Value' %}

sebastienros:
I looked into it since there is not such thing by default in Liquid, why there isn't and how people do without it. Found a straightforward solution. Here is an example:

```liquid
{% assign ages = people | map: 'age' | uniq %}
{%- for age in ages %}
  {%- assign people_by_age = people | where: 'age', age | sort: 'name' %}
  {%- for person in people_by_age %}
  {{ person.age }} {{ person.name }}
  {%- endfor%}
{%- endfor%}
```

[Playground](https://liquidjs.com/playground.html#eyUgYXNzaWduIGFnZXMgPSBwZW9wbGUgfCBtYXA6ICdhZ2UnIHwgdW5pcSAlfQ0KeyUtIGZvciBhZ2UgaW4gYWdlcyAlfQ0KICB7JS0gYXNzaWduIHBlb3BsZV9ieV9hZ2UgPSBwZW9wbGUgfCB3aGVyZTogJ2FnZScsIGFnZSB8IHNvcnQ6ICduYW1lJyAlfQ0KICB7JS0gZm9yIHBlcnNvbiBpbiBwZW9wbGVfYnlfYWdlICV9DQogIHt7IHBlcnNvbi5hZ2UgfX0ge3sgcGVyc29uLm5hbWUgfX0NCiAgeyUtIGVuZGZvciV9DQp7JS0gZW5kZm9yJX0NCg==,ewogICJwZW9wbGUiOiBbCiAgICB7ICJhZ2UiOiAxMiwgIm5hbWUiOiAiYm9iIiB9LAogICAgeyAiYWdlIjogMTUsICJuYW1lIjogImNhcm9sIiB9LAogICAgeyAiYWdlIjogMTIsICJuYW1lIjogImFsaWNlIiB9CiAgXQp9Cg==)

I am not opposed to adding special filters eithers, but this would have to be opt-in via the FluidOptions, and documented as non-standard.

yegorandrosov:
@sebastienros thanks a lot
i think indexer does not work with map filter? but i figured out that dot works

{% assign osVersions = model.Variants | map: 'CustomFields.97f9af79-a4ea-4766-8eac-b61da5d89de7.Value' | uniq %}

Issue #620, opened 2024-01-25
Where filter does not handle truthy values correctly


williamb1024:
When using the `where` filter with a single argument, the filter does not perform *truthy* comparisons. The equality comparison is performed with the property value controlling. 

https://github.com/sebastienros/fluid/blob/74cb7421a6400acb425fd7448552475f2f18178c/Fluid/Filters/ArrayFilters.cs#L156

Switching `itemValue` and `targetValue`, such that `targetValue` is controlling would result in *truthy* comparisons.

MoazAlkharfan:
Here's a workaround with adding a new filter named `twhere`

Usage `twhere` instead of `where`:
```
{% assign filteredItems = items | twhere: "property" %}
```

Register in the filters:
```c#
builder.AddFluid(x => {
    x.TemplateOptions.Filters.AddFilter(WhereFilter.Name, WhereFilter.Where);
});
```

Filter implementation with workaround:
```c#
internal static class WhereFilter
{
    internal const string Name = "twhere";

    // Track issue: https://github.com/sebastienros/fluid/issues/620
    internal static async ValueTask<FluidValue> Where(FluidValue input, FilterArguments arguments, TemplateContext context)
    {
        if (input.Type != FluidValues.Array)
        {
            return input;
        }

        // First argument is the property name to match
        var member = arguments.At(0).ToStringValue();

        // Second argument is the value to match, or 'true' if none is defined
        var targetValue = arguments.At(1).Or(BooleanValue.True);

        var list = new List<FluidValue>();

        foreach (var item in input.Enumerate(context))
        {
            var itemValue = await item.GetValueAsync(member, context);

            if (itemValue.Equals(targetValue) ||
                // Issue workaround for equality inconsistencies: to check if the other equals self
                targetValue.Equals(itemValue))
            {
                list.Add(item);
            }
        }

        return new ArrayValue(list);
    }
}
```

sebastienros:
So instead of comparing to `true` we actually need to handle the single argument separately and convert it using `ToBooleanValue()`. Can someone create a PR and a test for that?

gabbersepp:
@MoazAlkharfan I don't think that this is a correct fix.
According to the shopify liquid documentation, everything is truthy, except `nil` and `false` which also means, that I easily can use:

`where: "Name"` to find all items where the property `Name` has any value.

Am I right?

williamb1024:
The solution I came up with is:
``` CSharp
[...]

PR #707, opened 2024-10-24
Adjust Where filter to properly handle truthy comparisons.

williamb1024:
Corrects where to follow the `where` documentation from liquid language repository.  Additionally, it deals with a case where the comparison value is actually false, which would lead to the comparison being performed with a true value.

Corrects #620

### DIFF
--- a/Fluid.Tests/ArrayFiltersTests.cs
+++ b/Fluid.Tests/ArrayFiltersTests.cs
@@ -372,6 +372,34 @@ namespace Fluid.Tests
         }
 
         [Fact]
+        public async Task WhereWithTruthy()
+        {
+            var input = new ArrayValue(new[]
+            {
+                new ObjectValue(new { Title = "a", Pinned = true, Missing = 1 }),
+                new ObjectValue(new { Title = "b", Pinned = false }),
+                new ObjectValue(new { Title = "c", Pinned = true, Missing = 1 })
+            });
+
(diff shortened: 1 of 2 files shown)


sebastienros:
Can you add the failing case that triggered you filing the issue? Code looks good, I will have to debug to understand the differences ( don't remember the internals details)

Fluid.Tests/ArrayFiltersTests.cs
```diff
@@ -371,6 +371,55 @@ public async Task Where()
...
+                new ObjectValue(new { Title = "b", Pinned = false }),
+                new ObjectValue(new { Title = "c", Pinned = true, Missing = 1 })
+            });
+
+            var options = new TemplateOptions();
+            var context = new TemplateContext(options);
+            options.MemberAccessStrategy.Register(new { Title = "a", Pinned = true }.GetType());
+            options.MemberAccessStrategy.Register(new { Title = "a", Pinned = true, Missing = 1 }.GetType());
+
+            // x | where: "Missing"
```
sebastienros, line 389: What would be the behavior for `where: Pinned`, it's a boolean, I would expect 2. Can you add the test?

williamb1024: It is 2 as you expect. The `Where` test (above `WhereWithUndefined`) already performs that check.

PR #709, opened 2024-11-05
Mark FluidValue.WriteToAsync() as abstract

hishamco:
While I'm working on https://github.com/OrchardCMS/OrchardCore/pull/16956 I notice that subclass `FluidValue` forces me to implement `Write()` instead of `WriteAsync()` and a warning shows up because the first method is obsolete. A better option is to mark `WriteAsync()` as `abstract` so the default implement is async and `Write()` method is still valid for backward compatibility

### DIFF
--- a/Fluid/Values/FluidValue.cs
+++ b/Fluid/Values/FluidValue.cs
@@ -15,7 +15,7 @@ namespace Fluid.Values
             WriteToAsync(writer, encoder, cultureInfo).GetAwaiter().GetResult();
         }
 
-        public abstract ValueTask WriteToAsync(TextWriter writer, TextEncoder encoder, CultureInfo cultureInfo);
+        public virtual ValueTask WriteToAsync(TextWriter writer, TextEncoder encoder, CultureInfo cultureInfo);
 
         private static Dictionary<Type, Type> _genericDictionaryTypeCache = new();
 
--- a/Fluid/Values/FluidValue.cs
+++ b/Fluid/Values/FluidValue.cs
@@ -15,7 +15,7 @@ namespace Fluid.Values
             WriteToAsync(writer, encoder, cultureInfo).GetAwaiter().GetResult();
         }
 
-        public virtual ValueTask WriteToAsync(TextWriter writer, TextEncoder encoder, CultureInfo cultureInfo);
+        public virtual ValueTask WriteToAsync(TextWriter writer, TextEncoder encoder, CultureInfo cultureInfo) => default;
 
         private static Dictionary<Type, Type> _genericDictionaryTypeCache = new();
 
--- a/Fluid/Values/FluidValue.cs
+++ b/Fluid/Values/FluidValue.cs
@@ -12,7 +12,7 @@ namespace Fluid.Values
         [Obsolete("WriteTo is obsolete, prefer the WriteToAsync method.")]
         public virtual void WriteTo(TextWriter writer, TextEncoder encoder, CultureInfo cultureInfo)
         {
-            WriteToAsync(writer, encoder, cultureInfo);
+            WriteToAsync(writer, encoder, cultureInfo).GetAwaiter().GetResult();
         }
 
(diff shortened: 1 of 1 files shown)


Fluid/Values/FluidValue.cs
```diff
@@ -10,16 +10,13 @@ public abstract class FluidValue : IEquatable<FluidValue>
...
-        public abstract void WriteTo(TextWriter writer, TextEncoder encoder, CultureInfo cultureInfo);
-
-        public virtual ValueTask WriteToAsync(TextWriter writer, TextEncoder encoder, CultureInfo cultureInfo)
+        public virtual void WriteTo(TextWriter writer, TextEncoder encoder, CultureInfo cultureInfo)
         {
-#pragma warning disable CS0618 // Type or member is obsolete
-            WriteTo(writer, encoder, cultureInfo);
-#pragma warning restore CS0618 // Type or member is obsolete
-            return default;
+            WriteToAsync(writer, encoder, cultureInfo);
```
sebastienros, line 15:
```suggestion
            WriteToAsync(writer, encoder, cultureInfo).GetAwaiter().GetResult();
```

Fluid/Values/FluidValue.cs
```diff
@@ -10,16 +10,13 @@ public abstract class FluidValue : IEquatable<FluidValue>
...
         }
 
+        public abstract ValueTask WriteToAsync(TextWriter writer, TextEncoder encoder, CultureInfo cultureInfo);
```
sebastienros, line 18: Keep it virtual or this is a breaking change.

hishamco: I remembered you break many things in the past :)

Anyhow I will revert this

Fluid/Values/FluidValue.cs
```diff
@@ -10,16 +10,13 @@ public abstract class FluidValue : IEquatable<FluidValue>
...
+        public virtual void WriteTo(TextWriter writer, TextEncoder encoder, CultureInfo cultureInfo)
         {
-#pragma warning disable CS0618 // Type or member is obsolete
-            WriteTo(writer, encoder, cultureInfo);
-#pragma warning restore CS0618 // Type or member is obsolete
-            return default;
+            WriteToAsync(writer, encoder, cultureInfo).GetAwaiter().GetResult();
         }
 
+        public abstract ValueTask WriteToAsync(TextWriter writer, TextEncoder encoder, CultureInfo cultureInfo);
```
hishamco, line 18:
```suggestion
        public virtual ValueTask WriteToAsync(TextWriter writer, TextEncoder encoder, CultureInfo cultureInfo);
```

Fluid/Values/FluidValue.cs
```diff
@@ -10,16 +10,13 @@ public abstract class FluidValue : IEquatable<FluidValue>
...
-        public abstract void WriteTo(TextWriter writer, TextEncoder encoder, CultureInfo cultureInfo);
-
-        public virtual ValueTask WriteToAsync(TextWriter writer, TextEncoder encoder, CultureInfo cultureInfo)
+        public virtual void WriteTo(TextWriter writer, TextEncoder encoder, CultureInfo cultureInfo)
         {
-#pragma warning disable CS0618 // Type or member is obsolete
-            WriteTo(writer, encoder, cultureInfo);
-#pragma warning restore CS0618 // Type or member is obsolete
-            return default;
+            WriteToAsync(writer, encoder, cultureInfo).GetAwaiter().GetResult();
```
sebastienros, line 15: This is breaking everything. Current implementations of `WriteTo` that ignore the obsolete message or are already built have the code to run. That's what while [obsolete] is here the async one calls `WriteTo`. It's just if you implement `WriteToAsync` that it's picked up.

So it's fine to have `virtual void WriteTo` such that we don't HAVE to implement it, but that's the only change to make.

hishamco: That's work too

hishamco:
Closed in favor of #711

PR #710, opened 2024-11-06
Add direct support for structs

### LINKED ISSUE #556: Supporting `struct` by default, would be great

sebastienros:
Fixes #556


hishamco:
Can we create `2.12.0` release after merging this PR?

sebastienros:
@hishamco Before we do a new release I want to get someone's feedback.

hishamco:
Feedback on the latest changes?

PR #711, opened 2024-11-06
Mark WriteTo() virtual

hishamco:


hishamco:
@sebastienros according to your suggestion to mark the async method `virtual` not `abstract` you will notice that any `FluidValue` implement will no longer require to implement that method, that's why I suggested as the beginning to become `abstract` instead

PR #712, opened 2024-11-06
Improve base64 methods on .NET 6

sebastienros:


Fluid/Filters/MiscFilters.cs
```diff
@@ -106,7 +107,10 @@ public static ValueTask<FluidValue> Handleize(FluidValue input, FilterArguments
                 }
             }
 
-            static bool IsCapitalLetter(char c) => c >= 'A' && c <= 'Z';
+            static bool IsCapitalLetter(char c)
```
lahma, line 110: There might be a different with the `IsInRange` trick: https://sharplab.io/#v2:C4LglgNgNAJiDUAfAAgZgATIEzoMLoG8BYAKHXOQEYA2dAIwHsGJ0BJAZ1wEMAHMYLhAAyAU2DARAJwAUAYwAWXSelkBKdAF4AfCvRh26LRvQByAIIn0XAHYx0AHmMmAWiYDcpcphr0mLDtx8AsJiEpJYcorKslDoAK7s8gySwOgAtmDWsQlJKelcAB7q2ujScZnAqnLoALSmFuqOpeXWldIulnXmJqoeJAC+QA=

Fluid/Utils/HexUtilities.cs
```diff
@@ -0,0 +1,24 @@
+﻿#if NET6_0_OR_GREATER
+
+namespace Fluid.Utils;
+
+public class HexUtilities
```
lahma, line 5: maybe `internal static`?

sebastienros: I want to use it in the benchmark, and I can't get it to work because of public key (trust me I know how it works, but maybe I don't). I may just remove the benchmarks now that I have the result.

Fluid/Filters/MiscFilters.cs
sebastienros:
```
| Method | Job               | Runtime  | Mean     | Error     | StdDev   | Allocated |
|------- |------------------ |--------- |---------:|----------:|---------:|----------:|
| M1     | ShortRun-.NET 8.0 | .NET 8.0 | 15.90 ns |  4.675 ns | 0.256 ns |         - |
| M2     | ShortRun-.NET 8.0 | .NET 8.0 | 22.86 ns | 44.572 ns | 2.443 ns |         - |
| M1     | ShortRun-.NET 9.0 | .NET 9.0 | 16.24 ns |  6.305 ns | 0.346 ns |         - |
| M2     | ShortRun-.NET 9.0 | .NET 9.0 | 16.32 ns |  5.215 ns | 0.286 ns |         - |
```

```c#
    private string _letters = "01234567890abcdefghijklmnopqrstuvwxyZABCDEFGHIJKLMNOPQRSTUVWXYZ";

    [Benchmark]
    public void M1()
    {
        foreach (var c in _letters.AsSpan())
        {
            IsCapitalLetter(c);
        }
    }

    [Benchmark]
    public void M2()
    {
        foreach (var c in _letters.AsSpan())
        {
            IsCapitalLetter2(c);
        }
    }
```

lahma: Odd that it behaves badly on NET 8 🤷🏻‍♂️ thanks for checking out

sebastienros: If I remove the loop it's different ...

sebastienros:
```
| Method | Job               | Runtime  | Mean      | Error     | StdDev    | Median    | Allocated |
|------- |------------------ |--------- |----------:|----------:|----------:|----------:|----------:|
| M1     | ShortRun-.NET 8.0 | .NET 8.0 | 0.0943 ns | 2.9791 ns | 0.1633 ns | 0.0000 ns |         - |
| M2     | ShortRun-.NET 8.0 | .NET 8.0 | 0.0015 ns | 0.0466 ns | 0.0026 ns | 0.0000 ns |         - |
| M1     | ShortRun-.NET 9.0 | .NET 9.0 | 0.0053 ns | 0.1502 ns | 0.0082 ns | 0.0011 ns |         - |
| M2     | ShortRun-.NET 9.0 | .NET 9.0 | 0.0000 ns | 0.0000 ns | 0.0000 ns | 0.0000 ns |         - |
```

PR #718, opened 2024-11-22
Update .NET Core 9.0

hishamco:


hishamco:
@sebastienros  did I miss something here?

.github/workflows/build.yml
```diff
@@ -17,6 +17,9 @@ jobs:
 
     steps:
     - uses: actions/checkout@v4
+    - uses: actions/setup-dotnet@v4
+      with:
+        global-json-file: global.json
```
lahma, line 22: didn't know of this switch, cool

Fluid.MvcViewEngine/Fluid.MvcViewEngine.csproj
```diff
@@ -1,39 +1,43 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
-  <PropertyGroup>
```
lahma, line 3: Changing spaces to tabs doesn't feel right

Fluid/Fluid.csproj
```diff
@@ -1,53 +1,68 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
-  <PropertyGroup>
```
lahma, line 3: why change to tabs?

Fluid.Tests/Fluid.Tests.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>net6.0;net8.0</TargetFrameworks>
```
lahma, line 4: should probably drop net6.0 all around

.github/workflows/build.yml
hishamco: Me too 😄

lahma:
I think NET 9 SDK is a good addition, but I also believe that NET 9 target is unnecessary. SDKs ship with new clever ways to transform your code to more performant structures (when you use collection expression and friends for example).

Fluid.Tests/Fluid.Tests.csproj
hishamco: When I format the file in VS it uses tabs over spaces, regular tab is 4 spaces

lahma:
> > but I also believe that NET 9 target is unnecessary
> 
> Have you seen the framework references in the csproj files? But for `Fluid.csproj` I agree this is not necessary.

Yes for the consuming projects `net9.0` makes sense, just for publishing I see little value.

sebastienros:
MvcViewEngines is publishing, and has a reference to the aspnet runtime. I am afraid it would pull net8.0 aspnetcore if you used a net9.0 app. I can test it.

lahma (approved):

sebastienros:
I am still applying some changes, don't merge

lahma:
The latest release seems to cause some trouble.. https://www.nuget.org/packages/Fluid.Core lists 

image

Which might the cause for some problems with NJsonSchema upgrade:

https://github.com/RicoSuter/NJsonSchema/actions/runs/11999376115/job/33447213638?pr=1758

>   20:34:37 [DBG]    System.TypeInitializationException : The type initializer for 'LiquidTemplate' threw an exception.
>   20:34:37 [DBG] ---- System.IO.FileNotFoundException : Could not load file or assembly 'System.Text.Encodings.Web, Version=9.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'. The system cannot find the file specified.

sebastienros:
This was an intentional change which I thought would be fine ... going to rever this. The issue is with MEFA, right ?

sebastienros:
> There was a conflict between "System.Text.Encodings.Web, Version=8.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51" and "System.Text.Encodings.Web, Version=9.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51". 

That's the issue with bumping to 9.0.0, it's compatible with netstandard2.0, and there is no reason we shouldn't upgrade, but then other libraries might pull a different version directly and that could be an issue. Is NJsonSchema referencing System.Text.Encodings.Web, Version=8.0.0.0 directly?

sebastienros:
```
      └─ NJsonSchema (v11.0.0)
         └─ System.Text.Json (v8.0.5)
            └─ System.Text.Encodings.Web (v8.0.0)
```

sebastienros:
And it's only for net462, netstandard2.0, net472 since they have to reference STJ.

sebastienros:
FYI I used `dotnet nuget why .\NJsonSchema.sln System.Text.Encodings.Web` in case you were not aware of the command

lahma:
I think with general netstandard it would be best to require lowest possible. A weird problem to see when latest SDK is being used and it should automatically resolve dependencies and version forwards.

PR #728, opened 2024-12-06
Assigned delegate

deanebarker:
Allows the execution of a C# delegate whenever values are assigned via an `assign` or `capture` tag.

### DIFF
--- a/Fluid.Tests/AssignmentCaptureTests.cs
+++ b/Fluid.Tests/AssignmentCaptureTests.cs
@@ -1,0 +1,55 @@
+using System.Linq;
+using Fluid.Values;
+using Fluid.Filters;
+using Xunit;
+using System.Threading.Tasks;
+using Fluid.Tests.Extensibility;
+
+namespace Fluid.Tests
+{
+    public class AssignmentCaptureTests
+    {
+#if COMPILED
+        private static FluidParser _parser = new FluidParser(new FluidParserOptions { AllowFunctions = true }).Compile();
+#else
+        private static FluidParser _parser = new FluidParser(new FluidParserOptions { AllowFunctions = true });
+#endif
+
+
+        [Fact]
+        public async Task Assign()
+        {
+            string buffer = null;
+
+            var source = @"
+                {% assign a = 'b' %}
+            ";
+
+            _parser.TryParse(source, out var template, out var error);
+            var context = new TemplateContext
+            {
+                Assigned = (v) => buffer = v.ToStringValue()
+            };
+            var result = await template.RenderAsync(context);
+            Assert.Equal("b", buffer);
+        }
+
+        [Fact]
+        public async Task Capture()
+        {
+            string buffer = null;
+
+            var source = @"
+                {% capture a %}b{% endcapture %}
+            ";
+
+            _parser.TryParse(source, out var template, out var error);
+            var context = new TemplateContext
+            {
+                Assigned = (v) => buffer = v.ToStringValue()
+            };
+            var result = await template.RenderAsync(context);
+            Assert.Equal("b", buffer);
+        }
+    }
+}
--- a/Fluid.Tests/AssignmentCaptureTests.cs
+++ b/Fluid.Tests/AssignmentCaptureTests.cs
@@ -17,39 +17,18 @@ namespace Fluid.Tests
 
         [Fact]
         public async Task Assign()
-
         {
-            string buffer = null;
-
             var source = @"
-                {% assign a = 'b' %}
+                {%- assign a = 'b' %}{{a-}}
             ";
 
             _parser.TryParse(source, out var template, out var error);
             var context = new TemplateContext
             {
-                Assigned = (v) => buffer = v.ToStringValue()
+                Assigned = (identifier, value, context) => new StringValue(value.ToStringValue() + "_altered")
             };
             var result = await template.RenderAsync(context);
-            Assert.Equal("b", buffer);
-        }
-
-        [Fact]
-        public async Task Capture()
-        {
-            string buffer = null;
-
-            var source = @"
-                {% capture a %}b{% endcapture %}
-            ";
-
-            _parser.TryParse(source, out var template, out var error);
-            var context = new TemplateContext
-            {
-                Assigned = (v) => buffer = v.ToStringValue()
-            };
-            var result = await template.RenderAsync(context);
-            Assert.Equal("b", buffer);
+            Assert.Equal("b_altered", result);
         }
     }
 }
--- a/Fluid.Tests/AssignmentCaptureTests.cs
+++ b/Fluid.Tests/AssignmentCaptureTests.cs
@@ -15,9 +15,9 @@ namespace Fluid.Tests
         private static FluidParser _parser = new FluidParser(new FluidParserOptions { AllowFunctions = true });
 #endif
 
-
         [Fact]
         public async Task Assign()
+
         {
             string buffer = null;
 
--- a/Fluid/TemplateContext.cs
+++ b/Fluid/TemplateContext.cs
@@ -159,7 +159,6 @@ namespace Fluid
         /// </summary>
         public Func<string, string, ValueTask<string>> Captured { get; set; }
 
-
         /// <summary>
         /// Gets or sets the delegate to execute when a member has been assigned via capture or assign
         /// </summary>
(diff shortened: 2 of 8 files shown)


Fluid/TemplateContext.cs
```diff
@@ -156,7 +157,13 @@ public void IncrementSteps()
         /// <summary>
         /// Gets or sets the delegate to execute when a Capture tag has been evaluated.
         /// </summary>
-        public Func<string, string, ValueTask<string>> Captured { get; set; }
+        public Func<string, string, ValueTask<string>> Captured { get; set; }
+
+
+        /// <summary>
+        /// Gets or sets the delegate to execute when a member has been assigned via capture or assign
+        /// </summary>
```
hishamco, line 165:
```suggestion

        /// <summary>
        /// Gets or sets the delegate to execute when a member has been assigned via capture or assign
        /// </summary>
```

Fluid.Tests/AssignmentCaptureTests.cs
```diff
@@ -0,0 +1,55 @@
...
+    {
+#if COMPILED
+        private static FluidParser _parser = new FluidParser(new FluidParserOptions { AllowFunctions = true }).Compile();
+#else
+        private static FluidParser _parser = new FluidParser(new FluidParserOptions { AllowFunctions = true });
+#endif
+
+
+        [Fact]
+        public async Task Assign()
```
hishamco, line 20:
```suggestion
#endif

        [Fact]
        public async Task Assign()

```

hishamco (changes requested):

Fluid/TemplateContext.cs
hishamco: Remove unnecessary line

Fluid.Tests/AssignmentCaptureTests.cs
hishamco: Same here

hishamco (approved):

Fluid.Tests/AssignmentCaptureTests.cs
```diff
@@ -0,0 +1,55 @@
...
+            string buffer = null;
+
+            var source = @"
+                {% capture a %}b{% endcapture %}
+            ";
+
+            _parser.TryParse(source, out var template, out var error);
+            var context = new TemplateContext
+            {
+                Assigned = (v) => buffer = v.ToStringValue()
```
sebastienros, line 49: I'd rather separate both, or there is no way to just extend the assign tag only. If someone wants both then the same event can be added to both.

Fluid/TemplateContext.cs
```diff
@@ -1,274 +1,280 @@
...
+        /// <param name="modelNamesComparer">An optional <see cref="StringComparer"/> instance used when comparing model names.</param>
+        public TemplateContext(TemplateOptions options, StringComparer modelNamesComparer = null)
+        {
+            Options = options;
+            LocalScope = new Scope(options.Scope);
+            RootScope = LocalScope;
+            CultureInfo = options.CultureInfo;
+            TimeZone = options.TimeZone;
+            Captured = options.Captured;
+            Assigned = options.Assigned;
```
sebastienros, line 56: Good. It's important we can change the options globally or per context.

Fluid/TemplateContext.cs
```diff
@@ -1,274 +1,280 @@
...
+
+        /// <summary>
+        /// Gets or sets the delegate to execute when a Capture tag has been evaluated.
+        /// </summary>
+        public Func<string, string, ValueTask<string>> Captured { get; set; }
+
+        /// <summary>
+        /// Gets or sets the delegate to execute when a member has been assigned via capture or assign
+        /// </summary>
+        public Action<FluidValue> Assigned { get; set; }
```
sebastienros, line 165: Like `Captured` this should return a `ValueTask`. But unlike `Captured`, and that's something that could be changed at some point, we should also provide the `TemplateContext` as an argument because sometimes you need to take the options into account to do something. I have had to add it in the most recent commit I did, and obsolete the previous methods. So let's not miss it here.

Fluid.Tests/AssignmentCaptureTests.cs
deanebarker: How do want them named?

Fluid/TemplateContext.cs
deanebarker: So, should the delegate be `async`? (warning: async programming is a weak spot for me...)

sebastienros: What it means is the delegate signature needs to return a `ValueTask` so instead of `Action<FluidValue>` it becomes a `Func<FluidValue, ValueTask>`. Then to add `TemplateContext` as the input, it becomes `Func<FluidValue, TemplateContext, ValueTask>`.

I will need to help you though for the invokation since there is an early non-async exit (`Awaited` method) and I don't expect you to know how to handle that (you can try still).

deanebarker: ...so, I wasn't planning on do anything with the return value. My goal was just to get variables "out" of the template. Do you want something to return and be assigned inside the context?

sebastienros: Let me update your PR, you'll see. To make something "async" you just need to return "ValueTask<T>" or "Task<T>", you don't use it, the compiler/runtime will.

sebastienros:
PR updated. Also modified the Captured delegate with a breaking change, but makes it more powerful.

deanebarker:
@sebastienros To be totally clear, this is what I'm after. If you want to return the value to the context, that's fine, but goal for this PR was just to get them out.

```
  [Fact]
  public async Task Assign2()
  {
      var source = @"
          {%- assign a = 'b' %}
          {%- assign c = 'd' %}
      ";

      var dataAssignmentsFromTemplateExecution = new Dictionary<string, object>();

      _parser.TryParse(source, out var template, out var error);
      var context = new TemplateContext
      {
          Assigned = (identifier, value, context) => {
              dataAssignmentsFromTemplateExecution[identifier] = value.ToObjectValue();
              return value;
           }
      };
      var result = await template.RenderAsync(context);
      Assert.Equal("b", dataAssignmentsFromTemplateExecution["a"]);
      Assert.Equal("d", dataAssignmentsFromTemplateExecution["c"]);
      Assert.Equal(2, dataAssignmentsFromTemplateExecution.Count);
  }
```

sebastienros:
@deanebarker I understand, but since we are there I prefer to make it complete and not have someone ask for the feature late, with another breaking change.

deanebarker:
So my process of just passing the unmodified value back again is conceptually correct for what I want to do?

sebastienros:
Yes, that's the pattern.

deanebarker:
Great. Ship it.

sebastienros:
You could write your own extension method like this to have the same pattern as you had initially, while using the overly complicated solution I came up with:

```c#
public static TemplateContext OnAssigned(this TemplateContext this, Action<string, FluidValue> assigned))
{
    this.Assigned = (indentifier, value, context) => 
    {
        assigned(identifier, value);
        return ValueTask.FronmResult<FluidValue>(value);
    };
}
```

deanebarker:
You mean...complicate it further to avoid complication? ... 😀

sebastienros:
Every programmer should take some philosophy classes.

deanebarker:
"I think, therefore I am... _going to over-engineer this for no reason_."

Fluid/TemplateOptions.cs
```diff
@@ -1,93 +1,111 @@
...
+
+namespace Fluid
+{
+    public class TemplateOptions
+    {
+        /// <param name="identifier">The name of the property that is assigned.</param>
+        /// <param name="value">The value that is assigned.</param>
+        /// <param name="context">The <see cref="TemplateContext" /> instance used for rendering the template.</param>
+        /// <returns>The value which should be assigned to the property.</returns>
+        public delegate ValueTask<FluidValue> AssignedDelegate(string identifier, FluidValue value, TemplateContext context);
```
hishamco, line 14: I'm not sure why you split into two delegates and why the signature is the same

deanebarker: Honestly, I don't quite get this either. Whether you use assign or capture, it the same basic idea, conceptually.

PR #740, opened 2025-01-03
Add custom statements visitor support

### LINKED ISSUE #705: Introspection of parsed template using visitor
Hi,

I raised #578 and submitted a PR that received no comment and it looks like the advice now is to use the Visitor supports that's been recently added.

There is an example in the readme but it doesn't demonstrate what I need to do.

I need to walk the template and find instances of a given tag (in my case, `{% layout  '[id]' %}`). 

I want to be able to extract both the tag name (layout) and the input ID.

How do I do that using a Visitor?

Starting with a contrived example:

```
static void Main(string[] args)
{
    var template = new MyParser().Parse("{% layout 'layoutid' %}");
    var v = new Visitor();
    v.VisitTemplate(template);
}

public class Visitor : AstVisitor
{
    public override Statement Visit(Statement statement)
    {
        return base.Visit(statement);
    }
}
```

The `Statement` method breakpoints but I'm unable to access anything on the statement that is passed in.  The only method is `WriteToAsync` and there are no properties.

How do I extract the statement name and passed parameter?

Cheers

sebastienros:
Fixes #705

### DIFF
--- a/Fluid.Tests/VisitorTest.cs
+++ b/Fluid.Tests/VisitorTest.cs
@@ -1,5 +1,8 @@
-﻿using Fluid.Tests.Visitors;
+using Fluid.Ast;
+using Fluid.Tests.Visitors;
 using Fluid.Values;
+using Fluid.ViewEngine;
+using Parlot.Fluent;
 using Xunit;
 
 namespace Fluid.Tests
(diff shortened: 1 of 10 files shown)


sebastienros:
@knightmeister would you be able to verify it works for you?

knightmeister:
Sure, I can pick this up this evening - give me ~12h. Appreciated.

knightmeister:
@sebastienros apologies for not getting to this yesterday. Works perfectly now and as I'd expect - many thanks.

PR #755, opened 2025-02-12
ParentesesAfterFunctionName -> ParenthesesAfterFunctionName

hishamco:


sebastienros:
That should have broken something if it was used.

hishamco:
I will deprecate the old then

FYI I remembered there are many things broken over time without releasing a major release

PR #777, opened 2025-04-09
Ensure `ObjectValue.ToStringValue()` doesn’t return `null`

JanRomero:
Since [`ToString()` can return `null`](https://github.com/dotnet/coreclr/pull/23466), we must guard against this to prevent an `ArgumentNullException` later in `WriteToAsync()` when the string value gets encoded.

### DIFF
--- a/Fluid/Values/ObjectValueBase.cs
+++ b/Fluid/Values/ObjectValueBase.cs
@@ -177,7 +177,7 @@ namespace Fluid.Values
 
         public override string ToStringValue()
         {
-            return Convert.ToString(Value);
+            return Convert.ToString(Value) ?? "";
         }
 
         public override object ToObjectValue()


sebastienros:
Can you provide an example that would trigger such exception in a template?

JanRomero:
@sebastienros sure, this may be contrived but this’ll do it:

```csharp
public record struct MyModel(string Title, string? Value)
{
    public override string? ToString() => Value;
}
```

```csharp
var template = fluidParser.Parse("""
        Title: {{MyItem.Title}}
        Value: {{MyItem}}        
        """);

var dict = new Dictionary<string, object>() { ["MyItem"] = new MyModel() };

template.Render(new TemplateContext(dict)); //System.ArgumentNullException: "Value cannot be null. Arg_ParamName_Name"
```

Visual Studio will actually generate `string? ToString()` if nullable types are enabled for the project.

sebastienros:
I decided to accept null values but handle them, there might be other places that already return these.

https://github.com/sebastienros/fluid/pull/842

PR #778, opened 2025-04-09
Trigger “default” filter when value would result in empty output after `ToStringValue()`

JanRomero:
I have some model items that go like this, so they result in their value when they’re tagged without specifying a property:

```csharp
public class MyModel
{
    public string Value = "";
    public string Description = "Something something";
    public override string ToString() => this.Value;
}
```

In this configuration the default filter will see that the object is present and not display the default value, even though the object will later be rendered as an empty string.

```csharp
var template = fluidParser.Parse("{{ InterestingItem | default: '—' }}");

var dict = new Dictionary<string, object>() { ["InterestingItem"] = new MyModel() };

var html = template.Render(new TemplateContext(dict)); // html is "", but I would expect "—"
```

This PR adds a check for the object's actual string value, so the default filter will do its thing.

Cheers :)

### DIFF
--- a/Fluid/Filters/MiscFilters.cs
+++ b/Fluid/Filters/MiscFilters.cs
@@ -128,7 +128,7 @@ namespace Fluid.Filters
             }
             else
             {
-                if (input.IsNil() || input == BooleanValue.False || EmptyValue.Instance.Equals(input))
+                if (input.IsNil() || input == BooleanValue.False || EmptyValue.Instance.Equals(input) || input.ToStringValue() == "")
                 {
                     return arguments.At(0);
                 }


sebastienros:
I am wondering if the issue is that the `ToStringValue() -> ToString()` is actually invoked when rendering an object directly. I checked in Ruby AND Javascript implementations of Liquid it outputs the type name of the object.

If you want to provide a custom "empty/blank" behavior to your objects you should create a custom `FluidValue` type that will handle these cases.

I am not decided on the solution, will check how the Object.ToStringValue() should behave first.

JanRomero:
Thanks for having a look :)

C# also outputs the fully qualified type name (or for records just the name and the properties), unless `ToString()` is overridden.

I feel like the target audience for a template is end users who wouldn’t expect or want such internals to surface, so I try to make everything output something sensible. But I understand if you want to keep the current behaviour for compatibility with the original implementations.

PR #781, opened 2025-04-21
Avoid creating empty ArrayValue instances

lahma:
To my understanding, these ad hoc conversions required when doing interop between model and the template so these a transient and will not need to later reflect any state changes done in models. In NSwag there are properties that return empty enumerables with dynamic LINQ where outcome is an empty enumerator 90% of the cases.

### DIFF
--- a/Fluid/Values/FluidValue.cs
+++ b/Fluid/Values/FluidValue.cs
@@ -215,7 +215,7 @@ namespace Fluid.Values
                             return new DictionaryValue(new DictionaryDictionaryFluidIndexable(otherDictionary, options));
 
                         case FluidValue[] array:
-                            return new ArrayValue(array);
+                            return array.Length > 0 ? new ArrayValue(array) : ArrayValue.Empty;
                     }
 
                     // Check if it's a more specific IDictionary<string, V>, e.g. JObject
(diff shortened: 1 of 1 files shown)


Fluid/Values/FluidValue.cs
```diff
@@ -215,7 +215,7 @@ public static FluidValue Create(object value, TemplateOptions options)
                             return new DictionaryValue(new DictionaryDictionaryFluidIndexable(otherDictionary, options));
 
                         case FluidValue[] array:
-                            return new ArrayValue(array);
+                            return array.Length > 0 ? new ArrayValue(array) : ArrayValue.Empty;
```
hishamco, line 218:
```suggestion
                            return array.Length > 0
                                ? new ArrayValue(array)
                                : ArrayValue.Empty;
```

Fluid/Values/FluidValue.cs
```diff
@@ -264,20 +274,24 @@ public static FluidValue Create(object value, TemplateOptions options)
...
                         case IEnumerable enumerable:
-                            var fluidValues = new List<FluidValue>();
+                            List<FluidValue> fluidValues = null;
                             foreach (var item in enumerable)
                             {
+                                fluidValues ??= [];
                                 fluidValues.Add(Create(item, options));
                             }
-                            return new ArrayValue(fluidValues);
+                            return fluidValues != null ? new ArrayValue(fluidValues) : ArrayValue.Empty;
```
hishamco, line 283:
```suggestion

                            return fluidValues != null
                                ? new ArrayValue(fluidValues)
                                : ArrayValue.Empty;
```

hishamco (changes requested):

PR #782, opened 2025-04-21
Use single dictionary in DefaultMemberAccessStrategy

lahma:
It's wasteful do do a doble lookup to get onto type's member level. Now using single key. Added also simple `Lock` shim for non NET  9 targets.


sebastienros:
Check what I did recently, almost identical. Currently refactoring the whole thing for 3.0 so all members of the classes you pass will be accessible by default.

https://github.com/sebastienros/fluid/compare/main...sebros/memberstrategy#diff-11accc50705ae6f28b4b24b0ea6ab4de5c19ee01ae37dc7084aaca29f03d0e20R9

PR #786, opened 2025-04-25
Fix caching conflicts

sebastienros:


lahma:
This is interesting...

hishamco:
How does it fix the caching conflicts?!!

sebastienros:
I am iterating on this PR, I got a bug report, so first trying to see if I can repro without changes but different OSes, then I will add a test, and then the fix. Using this PR as the way to run on the oses I don't have (macos).

.github/workflows/pr.yml
```diff
@@ -12,9 +12,9 @@ jobs:
   build:
 
     strategy:
-      fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
+      fail-fast: false
```
lahma, line 17: This is why we generate GH Actions YAML with build tools like NUKE 😉

sebastienros: I used Copilot, it's even better ;)

Fluid/DefaultTemplateCache.cs
```diff
@@ -10,25 +11,28 @@ namespace Fluid;
...
+
+        _cache = new(RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? StringComparer.OrdinalIgnoreCase : StringComparer.Ordinal);
     }
 
     public bool TryGetTemplate(IFileInfo fileInfo, out IFluidTemplate template)
     {
         template = null;
 
-        if (_cache.TryGetValue(fileInfo.Name, out var cachedTemplate))
+        if (_cache.TryGetValue(fileInfo.PhysicalPath, out var cachedTemplate))
```
rynowak, line 30: Physical path can be null for in-memory filesystems. I think this will break people unless you require this at some other level.

sebastienros: This is problematic because `IFileInfo` is very limited here:

```c#
        //
        // Summary:
        //     The name of the file or directory, not including any path.
        string Name { get; }
        //
        // Summary:
        //     The path to the file, including the file name. Return null if the file is not
        //     directly accessible.
        string? PhysicalPath { get; }
```

Fluid/DefaultTemplateCache.cs
```diff
@@ -23,33 +22,31 @@ public TemplateCache()
         _cache = new(RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? StringComparer.OrdinalIgnoreCase : StringComparer.Ordinal);
     }
 
-    public bool TryGetTemplate(IFileInfo fileInfo, out IFluidTemplate template)
+    public bool TryGetTemplate(string subpath, DateTimeOffset lastModified, out IFluidTemplate template)
```
rynowak, line 25: 👍

PR #787, opened 2025-04-27
Reduce value type boxing in interop calls

lahma:
* special case common bool and int property access
* use abstract class instead of interface do reduce virtual dispatch cost

When profiling NJsonSchema, found out that now after optimizations almost 10% of allocations originate from boolean value boxing (very common in code generation templates).

image

### DIFF
--- a/Fluid/Accessors/PropertyInfoAccessor.cs
+++ b/Fluid/Accessors/PropertyInfoAccessor.cs
@@ -12,7 +12,7 @@ namespace Fluid.Accessors
         {
             Delegate d;
 
-            if (!propertyInfo.DeclaringType.IsValueType)
+            if (!propertyInfo.DeclaringType?.IsValueType == true)
             {
                 var delegateType = typeof(Func<,>).MakeGenericType(propertyInfo.DeclaringType, propertyInfo.PropertyType);
                 d = propertyInfo.GetGetMethod().CreateDelegate(delegateType);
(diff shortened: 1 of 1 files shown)


Fluid/Accessors/PropertyInfoAccessor.cs
```diff
@@ -30,8 +31,21 @@ public PropertyInfoAccessor(PropertyInfo propertyInfo)
...
+            else if (propertyInfo.PropertyType == typeof(int))
+            {
+                var invokerType = typeof(Int32Invoker<>).MakeGenericType(propertyInfo.DeclaringType);
+                _invoker = (Invoker) Activator.CreateInstance(invokerType, [d]);
+            }
+            else
+            {
+                var invokerType = typeof(Invoker<,>).MakeGenericType(propertyInfo.DeclaringType, propertyInfo.PropertyType);
+                _invoker = (Invoker) Activator.CreateInstance(invokerType, [d]);
+            }
```
hishamco, line 48:
```suggestion
            Type invokerType = null;
            if (propertyInfo.PropertyType == typeof(bool))
            {
                invokerType = typeof(BooleanInvoker<>).MakeGenericType(propertyInfo.DeclaringType);
            }
            else if (propertyInfo.PropertyType == typeof(int))
            {
                invokerType = typeof(Int32Invoker<>).MakeGenericType(propertyInfo.DeclaringType);
            }
            else
            {
                invokerType = typeof(Invoker<,>).MakeGenericType(propertyInfo.DeclaringType, propertyInfo.PropertyType);
            }
            
           _invoker = (Invoker) Activator.CreateInstance(invokerType, [d]);
```

Fluid/Accessors/PropertyInfoAccessor.cs
```diff
@@ -78,10 +92,41 @@ public Invoker(Delegate d)
...
         }
+
+        private sealed class BooleanInvoker<T> : Invoker
+        {
+            private readonly Func<T, bool> _d;
+
+            public BooleanInvoker(Delegate d)
+            {
+                _d = (Func<T, bool>)d;
+            }
```
hishamco, line 108:
```suggestion
            public BooleanInvoker(Delegate d) =>  _delegate = (Func<T, bool>)d;
```

Fluid/Accessors/PropertyInfoAccessor.cs
```diff
@@ -78,10 +92,41 @@ public Invoker(Delegate d)
...
+
+            public override object Invoke(object target)
+            {
+                return _d((T)target) ? BooleanValue.True : BooleanValue.False;
+            }
```
hishamco, line 113:
```suggestion
            public override object Invoke(object target) => _delegate((T)target) ? BooleanValue.True : BooleanValue.False;
```

Fluid/Accessors/PropertyInfoAccessor.cs
```diff
@@ -78,10 +92,41 @@ public Invoker(Delegate d)
...
-            public object Invoke(object target)
+            public override object Invoke(object target)
             {
                 return _d((T)target);
             }
         }
+
+        private sealed class BooleanInvoker<T> : Invoker
+        {
+            private readonly Func<T, bool> _d;
```
hishamco, line 103:
```suggestion
            private readonly Func<T, bool> _delegate;
```

Fluid/Accessors/PropertyInfoAccessor.cs
```diff
@@ -78,10 +92,41 @@ public Invoker(Delegate d)
...
+            public Int32Invoker(Delegate d)
+            {
+                _d = (Func<T, int>)d;
+            }
+
+            public override object Invoke(object target)
+            {
+                var value = _d((T)target);
+                return NumberValue.Create(value);
+            }
```
hishamco, line 129:
```suggestion
            private readonly Func<T, int> _delegate;

            public Int32Invoker(Delegate d) => _delegate = (Func<T, int>)d;

            public override object Invoke(object target) => NumberValue.Create(_delegate((T)target));
```

hishamco (changes requested):

sebastienros:
Did you check what I did in the other branch I mentioned last time? Similar to skip the boxing, but did it for any common type. Not the new specific classes though.

sebastienros:
No, it's a big breaking change, the goal is to ship a 3.0 with it. I'll move your improvements there though.

PR #788, opened 2025-04-27
Improve MemberExpression performance

lahma:
* use array without IReadOnlyList dispatch cost

### DIFF
--- a/Fluid/Ast/MemberExpression.cs
+++ b/Fluid/Ast/MemberExpression.cs
@@ -4,28 +4,33 @@ namespace Fluid.Ast
 {
     public sealed class MemberExpression : Expression
     {
-        public MemberExpression(MemberSegment segment)
+        private readonly MemberSegment[] _segments;
+
+        public MemberExpression(MemberSegment segment) : this([segment])
         {
-            Segments = [segment];
         }
 
-        public MemberExpression(IReadOnlyList<MemberSegment> segments)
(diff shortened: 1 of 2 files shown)


Fluid/Ast/MemberExpression.cs
```diff
@@ -4,28 +4,33 @@ namespace Fluid.Ast
...
+        internal MemberExpression(MemberSegment[] segments)
         {
-            Segments = segments ?? [];
+            _segments = segments ?? [];
 
-            if (Segments.Count == 0)
+            if (_segments.Length == 0)
             {
-                throw new ArgumentException("At least one segment is required in a MemberExpression");
+                ExceptionHelper.ThrowArgumentNullException(nameof(segments), "At least one segment is required in a MemberExpression");
```
hishamco, line 23:
```suggestion
                ArgumentNullException.ThrowIfNull(nameof(segments));
```

lahma: Only available from NET 7 onwards.

hishamco: Right, I forgot that the library targets multiple TFM, but we could use `ArgumentNullException.ThrowIfNull` in the helper you created if the TFM >= .NET 7

lahma: I don't understand the need for such complexity for a case that should never even be hit.

hishamco: Leave it then

PR #794, opened 2025-05-03
Fix liquid tag completions support

### LINKED ISSUE #793: Possible problem with 'continue' in a 'for' wrapped 'liquid' statement.
There might be an issue with the `continue` (and possibly `break`) when a `for` loop wraps a `liquid` tag.  This pattern is sometimes used in templates by a few people on our team.  [Note: We know it does work when the `liquid` tag is outside, wrapping the `for` loop.]

I verified this works on a few other Liquid sandboxes, but I included an example from the liquidjs.com sandbox below.

**Template**
```Liquid
{%- for i in (1..5) %}
    {%- liquid 
        if i > 3
            continue
        endif
        echo i
    %}
{%- endfor %}
```

**Expected Output** [Liquid Sandbox Example](https://liquidjs.com/playground.html#eyUtIGZvciBpIGluICgxLi41KSAlfQ0KICAgIHslLSBsaXF1aWQgDQogICAgICAgIGlmIGkgPiAzDQogICAgICAgICAgICBjb250aW51ZQ0KICAgICAgICBlbmRpZg0KICAgICAgICBlY2hvIGkNCiAgICAlfQ0KeyUtIGVuZGZvciAlfQ==,e30=)

```
123
```

**Actual Output (Fluid v2.24)**
```
12345
```

**Test**
```CSharp
[Fact]
public void ShouldContinueForLoopWithLiquidTag()
{
    var source = @"
        {%- for i in (1..5) %}
            {%- liquid 
                if i > 3
                    continue
                endif
                echo i
            %}
        {%- endfor %}
        ";

    var parser = new FluidParser( new FluidParserOptions { AllowLiquidTag = true } );
    Assert.True( parser.TryParse( source, out var template, out var errors ), errors );
    var rendered = template.Render();
    Assert.DoesNotContain( "45", rendered );
}
```

If I were to make a guess, it _might be_ that [the LiquidStatement might need to capture the result of the statement.WriteToAsync](https://github.com/sebastienros/fluid/blob/f3af1b37293e287abf0e4cc52db72b28dd823f8f/Fluid/Ast/LiquidStatement.cs#L18) and break if not `Completion.Normal`.

sebastienros:
Fixes #793

### DIFF
--- a/Fluid.Tests/ParserTests.cs
+++ b/Fluid.Tests/ParserTests.cs
@@ -994,21 +994,25 @@ class  {
         }
 
         [Fact]
-        public void ShouldParseLiquidTagWithBlocks()
+        public void LiquidTagShouldBreakOnCompletion()
         {
-            var source = @"
-{% liquid assign cool = true
-   if cool
-     echo 'welcome to the liquid tag' | upcase
-   endif 
-%}
-";
+            var source = """
+                {%- for i in (1..5) %}
+                    {%- liquid 
+                        if i > 3
+                            continue
+                        endif
+                        echo i
+                    %}
+                {%- endfor %}
+                """;
 
             var parser = new FluidParser(new FluidParserOptions { AllowLiquidTag = true });
             Assert.True(parser.TryParse(source, out var template, out var errors), errors);
             var rendered = template.Render();
-            Assert.Contains("WELCOME TO THE LIQUID TAG", rendered);
+            Assert.DoesNotContain("45", rendered);
         }
+         
 
         [Fact]
         public void ShouldParseFunctionCall()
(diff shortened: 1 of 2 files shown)


Fluid.Tests/ParserTests.cs
```diff
@@ -994,21 +994,25 @@ public void ShouldParseLiquidTag()
...
         {
-            var source = @"
-{% liquid assign cool = true
-   if cool
-     echo 'welcome to the liquid tag' | upcase
-   endif 
-%}
-";
+            var source = """
+                {%- for i in (1..5) %}
```
hishamco, line 1000: Is `-` after the percentage symbol for a reason? This is the first time I see it

sebastienros: In liquid it means that the spaces before the tag should not be rendered. Works on every tags or output tag.

PR #798, opened 2025-05-17
Add find/has tests

sebastienros:


sebastienros:
@lahma any idea why the build shows warnings about FileSystemAbstraction package 9.0 not compatible with net6.0 while there is a specific version for it, and even the public package references the correct version? Also try `dotnet nuget why` and it shows net6.0 uses FSA 9.0.

lahma:
> @lahma any idea why the build shows warnings about FileSystemAbstraction package 9.0 not compatible with net6.0 while there is a specific version for it, and even the public package references the correct version? Also try `dotnet nuget why` and it shows net6.0 uses FSA 9.0.

I've got a solution 😉  https://github.com/sebastienros/fluid/pull/799

Basically I believe they've installed hooks to packages that report problems on such obsolete platforms.

sebastienros:
The hooks are in the more recent version. They should not be used based on the config. I will ask someone from the team... Not disagreeing with dropping net6.0, will check.

PR #802, opened 2025-05-19
Fix scope case sensitivity

sebastienros:


Fluid/Scope.cs
```diff
@@ -1,33 +1,34 @@
-﻿using System.Runtime.CompilerServices;
 using Fluid.Values;
+using System.Runtime.CompilerServices;
```
hishamco, line 2: Why?!!

hishamco (approved):

Fluid/Scope.cs
sebastienros: VS moved it?

PR #814, opened 2025-08-07
Add `WithColorFilters` to `TemplateOptions` initialization

gustavorosolem:
The color filters were implemented but never added to the TemplateOptions.
I tried to find a reason why it wasn’t done, but couldn’t find any mention of it, so I thought it was just forgotten.

### DIFF
--- a/Fluid/TemplateOptions.cs
+++ b/Fluid/TemplateOptions.cs
@@ -120,7 +120,9 @@ namespace Fluid
             Filters.WithArrayFilters()
                 .WithStringFilters()
                 .WithNumberFilters()
-                .WithMiscFilters();
+                .WithMiscFilters()
+                .WithColorFilters();
         }
     }
 }
+


Fluid/TemplateOptions.cs
```diff
@@ -120,7 +120,9 @@ public TemplateOptions()
             Filters.WithArrayFilters()
                 .WithStringFilters()
                 .WithNumberFilters()
-                .WithMiscFilters();
+                .WithMiscFilters()
+                .WithColorFilters();
```
hishamco, line 124: Nice catch, seems I forgot to add it before

hishamco (approved):

hishamco:
Thanks for your contribution @gustavorosolem

Fluid/TemplateOptions.cs
sebastienros: I think it's actually on purpose because these are not part of the liquid spec and such are usually only opt in. What could be missing is a doc section listing all the With methods and special filters.

hishamco: https://shopify.dev/docs/api/liquid/filters/color_to_hex

sebastienros: This is Shopify specific. Like we have custom filters in Orchard. True Liquid reference is taken from https://github.com/Shopify/liquid. Anything that is in this repo and documented in Shopify is fine to use though.

hishamco: Unfortunately, we did not mention how to create a custom filter(s). We could improve the docs

hishamco: I got it, I will revert the latest commit, then write clear docs about it

PR #815, opened 2025-08-07
Revert "Add `WithColorFilters` to `TemplateOptions` initialization

hishamco:


hishamco:
@gustavorosolem I will revert the latest commit, and we need to make this clear in the docs. You can follow the convo [here](https://github.com/sebastienros/fluid/pull/814#discussion_r2260695645)

PR #847, opened 2025-11-03
Replace JsonWriterOptions with JsonSerializationOptions

sebastienros:


hishamco:
If this is not urgent to merge, I need to have a look tonight

PR #850, opened 2025-11-06
Implement EnumerateAsync

sebastienros:


hishamco:
This will introduce a breaking change, unless we provide backward compatibility

sebastienros:
@hishamco 3.0

hishamco:
@sebastienros, is there anything else we need to add to `3.0`? It's my pleasure to contribute

Issue #879, opened 2025-11-22
Create StrictVariables and StrictFilters


sebastienros:
For rendering the strict_variables and strict_filters will throw an exception if any unknown variables or filters are used.

Issue #849, opened 2025-11-04
Add an option to throw when undefined variables are used


sebastienros:
Based on the discussion in [this PR](https://github.com/sebastienros/fluid/pull/844):

@rodion-m wrote:

what do you think if we add an option like ThrowOnUndefinedVariable that will just init Undefined delegate like you showed? I really believe that it's important to have such an option from a UX perspective.

If a user set both these options - we can throw an exception.

Or an alternative:

Just add a descendant StrictTemplateOptions without Undefined delegate all (it should be preset). This solution will be more type safe.

sebastienros:
Related issue I just created after looking at SopifyLiquid options: https://github.com/sebastienros/fluid/issues/878

PR #880, opened 2025-11-22
Add StrictVariables and StrictFilters options

sebastienros:
Fixes https://github.com/sebastienros/fluid/issues/879
Fixes https://github.com/sebastienros/fluid/issues/849


sebastienros:
@rodion-m maybe not exactly what you want but if you need all the variable names in one shot then you can still use the Undefined callback.

PR #901, opened 2025-12-27
Source generation support

sebastienros:
AI generated, using GPT 5.2 with two prompts, one to decribe source generation integration in statements, and then another one to ask for a source generator.

I haven't even tried it yet, it's just crazy it managed to build something with absolutely no feedback in a single short (two actually since I split the requirements in two steps). Now I'll check if that works well, and what to improve.

Goals:
- Make startup faster as the parser doesn't have to be initialized (Parlot grammar) and the files don't need to be parsed either.
- Make rendering faster by removing the statements evaluation (virtual calls), and optimizing the generated code (don't render for-loop state, use non-async when possible, ...)


hishamco:
Before I have a look at it. Where are the Source Generators useful in Fluid?

sebastienros:
@hishamco I updated the main comment with the intended advantages

PR #909, opened 2026-01-20
Add Golden Liquid test suite

sebastienros:


sebastienros:
@jg-rp what's interesting for you is probably the list of tests that I had to skip, for good or bad reasons https://github.com/sebastienros/fluid/pull/909/changes#diff-555225c1c9a46c85711cd81838627218ddc83ecdaf7e127298300ccbe3b00297R27

PR #963, opened 2026-07-22
Add money filters

### LINKED ISSUE #238: Implement Money filters
https://help.shopify.com/en/themes/liquid/filters/money-filters
https://help.shopify.com/en/manual/payments/currency-formatting

Moved from Orchard Core issue.

lahma:
Fixes #238

Implements the [Shopify money filters](https://shopify.dev/docs/api/liquid/filters/money), registered with the opt-in `FilterCollection.WithMoneyFilters()` (like `WithColorFilters()`, they are not added by the `TemplateOptions` constructor):

| Filter | Source | Result |
| --- | --- | --- |
| `money` | `{{ 1134.65 \| money }}` | `$1,134.65` |
| `money_with_currency` | `{{ 1134.65 \| money_with_currency }}` | `$1,134.65 USD` |
| `money_without_currency` | `{{ 1134.65 \| money_without_currency }}` | `1,134.65` |
| `money_without_trailing_zeros` | `{{ 10.00 \| money_without_trailing_zeros }}` | `$10` |

### Configuration

A new `TemplateOptions.MoneyOptions`, mirrored on `TemplateContext` so a currency can be selected per rendering:

- **Culture based by default.** The currency and the separators come from `TemplateOptions.CultureInfo`. The invariant culture has no region, so `USD`/`$` is used, which matches Shopify's default `${{amount}}` format.
- **`Currency`**, an ISO 4217 code, also accepted as a filter argument: `{{ 10 | money: 'EUR' }}` and `{{ 10 | money: currency: 'EUR' }}`. Symbols and decimal digits come from an extensible `Currencies` collection seeded with the common codes; an unregistered code is rendered using the code itself as the symbol.
- **`AmountsInCents`** (default `false`) divides the input by 100, for templates ported from Shopify where prices are stored as integers.
- **`MoneyFormat` / `MoneyWithCurrencyFormat`** override the culture based formatting with the [Shopify currency formatting](https://help.shopify.com/en/manual/payments/currency-formatting) placeholders (`{{amount}}`, `{{amount_with_comma_separator}}`, …), plus `{{currency}}` and `{{currency_symbol}}` so a single format works with a currency resolved at rendering time. Formats are parsed once, when assigned.

### Notes

Amounts default to plain decimals rather than cents. The issue comment on #238 is a good illustration of what people expect: `{{ item.Price | money_without_currency }}` currently falls through unfiltered and renders `31.300` or `31.30` depending on the scale of the `decimal` that came out of the database. Money formatting pins the decimal count, and there is a test for it.

Two deliberate deviations from raw .NET currency formatting, both overridable with a format string:

- Amounts are rounded away from zero, as Shopify does, instead of .NET's banker's rounding.
- Negative amounts render as `-$10.00` rather than the accounting `($10.00)` that `"C"` produces for en-US and the invariant culture.

`README.md` has a new "Money filters" section, and `CLAUDE.md` is added.

### Testing

70 new tests in `Fluid.Tests/MoneyFiltersTests.cs`. Full suite passes on both the regular and the `Compiled=true` pass, and the build is clean on all four TFMs including netstandard2.0.


hishamco:
Just a little note before reviewing why we simplify this by having one `money` filter, then use arguments to implement the others

sebastienros:
AGENTS.md instead?

lahma:
> AGENTS.md instead?

Merged into AGENTS.md

PR #964, opened 2026-07-23
Cache member accessors, filters and small scopes during rendering

lahma:
Caches the things rendering re-resolves on every loop iteration and every render of a cached template, and stops `for` from rebuilding constants per execution.

BenchmarkDotNet, .NET 10 / AMD 5950X, against `df733e0`. Each side rebuilt from a clean `obj/bin` and verified by assembly marker, sides alternating, and any run the machine wasn't idle for discarded. `main` got the benefit of more samples than `branch`, so these deltas are conservative.

## Render

| Method | main | branch | |
|---|---:|---:|---:|
| Render | 32.80 μs | **23.50 μs** | **−28.4%** |
| Render (compiled parser) | 30.85 μs | **23.10 μs** | **−25.1%** |
| ParseAndRender | 35.81 μs | **27.73 μs** | **−22.6%** |
| Render allocations | 41.12 KB | **37.37 KB** | −9.1% |

## Shapes the catalogue template doesn't reach

`RenderScenarioBenchmarks` is new — heterogeneous collections, deep scope chains, scopes wider than the inline slots, names needing case conversion, numbers that can't be interned. These are where the caching could have *cost*.

| Benchmark | main | branch | | alloc |
|---|---:|---:|---:|---:|
| NestedLoopsWithParentLoop | 15.50 μs | **9.91 μs** | **−36.1%** | −52.1% |
| WideScope | 4.10 μs | **2.75 μs** | **−32.9%** | −20.9% |
| DeepScopeChain | 91.69 μs | **65.11 μs** | **−29.0%** | **−61.6%** |
| MonomorphicMembers | 7.61 μs | **5.73 μs** | **−24.7%** | −7.6% |
| PascalCaseMemberNames | 7.45 μs | **5.67 μs** | **−23.8%** | −9.0% |
| PolymorphicMembers | 8.74 μs | **6.88 μs** | **−21.3%** | −7.6% |
| NonInternedNumbers | 12.94 μs | **11.82 μs** | **−8.6%** | +1.9% |

Nested loops gain most: inline slots remove a `Dictionary` allocation per scope, and a nested loop creates a scope per iteration of the enclosing one.

## Cost

Parsing allocates **7–9% more** (`Parse` 3.21 → 3.44 KB): `IdentifierSegment` carries two more references, multiplied by parser backtracking. Parse time is +0.2% to +5.7% depending on the benchmark, i.e. at or near noise. Templates are parsed once and cached.

## Changes

- **`IdentifierSegment`** caches the resolved `IMemberAccessor` per call site, keyed on runtime type, comparer and a strategy-supplied token. A site that keeps missing — a heterogeneous loop, or a template rendered against several `TemplateOptions` — stops caching rather than allocating a replacement entry per access.
- **`MemberAccessStrategy.AccessorCacheToken`** — new, defaults to `null` (no caching) so custom strategies stay correct. `DefaultMemberAccessStrategy` retires its token only when a caller registers an accessor, so resolving a member doesn't invalidate every call site in the process. Only the exact type opts in: a subclass overriding `GetAccessor` may resolve from a source the strategy can't see.
- **`FilterExpression`** skips the async state machine when input and filter both complete synchronously, caches the resolved delegate, and gives up caching at a site that keeps missing. Errors still arrive as a faulted task.
- **`Scope`** holds up to 3 names in inline slots instead of a `Dictionary` — what a `for` loop needs (loop variable, `forloop`, `parentloop`). `Ordinal` and `OrdinalIgnoreCase` are compared without a virtual call.
- **`NumberValue`** shares instances for small non-negative whole numbers and renders them from precomputed text. Scale is part of a number's identity in Liquid, so only scale 0 qualifies; limited to targets where reading the scale is free.
- **`FluidValue.Create`** remembers types that are plain objects, skipping the enum check, type-code lookup and dozen pattern matches per item.
- **`MemberNameStrategies`** settles ordinally equal names without `ConvertName`, which allocates.
- **`ForStatement`** precomputes the `offset: continue` key and whitespace-only body check instead of rebuilding them with LINQ and `string.Join` per execution.

## Profile

CPU profile (ETW, 8 kHz) of the catalogue render, main → branch:

| | main | branch |
|---|---:|---:|
| `ObjectValueBase.GetValueAsync` | 10.9% | not in top 30 |
| `FilterExpression` async state machine | 4.8% | gone (sync path) |
| `TryFormatDecimal` + `NumberToString` | 2.9% | gone |
| `SpanHelpers.SequenceEqual` | 3.6% | 2.2% |

Remaining allocation is ~80% the final `ToString()` in `Render()`, i.e. inherent to returning a string.

## Tests

`RenderCacheInvalidationTests` (19 cases) covers what the caches could get wrong: accessors re-registered between renders; a strategy subclass that must never be cached; one template alternating across two `TemplateOptions`, including that it stops allocating once it gives up; one call site seeing alternating types; filters added, replaced and removed; `StrictFilters` and a throwing filter both arriving as faulted tasks; a null scope name throwing at the offending call; culture-invariance at the interning boundary; decimal-scale preservation. Suite: 2602 passing.

## Review notes

- `AccessorCacheToken` is new `protected internal virtual` surface — additive and binary-compatible, default disables caching.
- `Scope.Properties` now returns a snapshot rather than a live key view.
- `Fluid.Benchmarks/Program.cs` gains a `profile` mode: a steady-state loop for sampling profilers.

### DIFF
--- a/Fluid.Benchmarks/Program.cs
+++ b/Fluid.Benchmarks/Program.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Diagnostics;
 using BenchmarkDotNet.Running;
 
 namespace Fluid.Benchmarks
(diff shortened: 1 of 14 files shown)


sebastienros:
Too fast

lahma:
Ready now, I don't know of plans but could also be back-ported to 2.x if makes sense.

PR #979, opened 2026-08-19
Reduce render argument allocations

sebastienros:
The scope-management rewrite merged in #981 already removes the redundant render scope allocation. Named render arguments still allocate a temporary `KeyValuePair[]` on every execution, even though most calls pass only one or two arguments.

## Approach

- Keep one- and two-argument values inline without temporary arrays.
- Rent cleared `FluidValue[]` storage only for three or more arguments and return it on both success and failure.
- Continue evaluating every expression and `Assigned` callback in the caller scope before publishing any argument into the isolated render scope.
- Add callback-order/error-cleanup coverage and a focused two-argument BenchmarkDotNet case.

## Benchmark

BenchmarkDotNet 0.15.8, .NET 10.0.11 Arm64, Apple M4 Pro, in-process toolchain, 3 warmup and 5 measurement iterations:

| Shape | `origin/main` | This change | Difference |
|---|---:|---:|---:|
| Two named render arguments | 207.2 ns / 744 B | 212.1 ns / 688 B | **-56 B (-7.5%)** |

The timing confidence intervals overlap (198.3-216.1 ns baseline and 207.8-216.5 ns after), so this is treated as allocation-neutral throughput rather than a speed improvement. A pool-only two-argument version measured 218.5 ns / 688 B; inline storage was retained for the common case, with pooling reserved for larger argument sets.

## Validation

- Targeted interpreted and compiled render tests: 3 passed in each mode
- `dotnet test --no-restore`: 2,750 passed, 20 skipped, 0 failed
- `dotnet test --no-restore /p:Compiled=true`: 2,750 passed, 20 skipped, 0 failed
- Release build of `Fluid`: netstandard2.0, net8.0, net9.0, and net10.0 with no warnings or errors

This PR now targets `main` directly because #981 merged the preceding scope-management layer.

### DIFF
--- a/Fluid.Benchmarks/RenderScenarioBenchmarks.cs
+++ b/Fluid.Benchmarks/RenderScenarioBenchmarks.cs
@@ -32,6 +32,7 @@ namespace Fluid.Benchmarks
         private readonly IFluidTemplate _pascalCaseTemplate;
         private readonly IFluidTemplate _decimalPriceTemplate;
         private readonly IFluidTemplate _renderScopeTemplate;
+        private readonly IFluidTemplate _renderArgumentsTemplate;
         private readonly IFluidTemplate _includeScopeTemplate;
 
         private readonly List<object> _mixedItems = new(ItemCount);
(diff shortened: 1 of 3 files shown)


sebastienros:
Closing because this performance work is investigatory only; commit caaeb4f remains available on the branch for traceability.

